Issue #41, opened 2026-04-24
Staging transaction id collision under stamina retry can silently drop chunk ingestion


JasperHG90:
## Summary

Two concurrent batch jobs can end up passing the same id into `AsyncTransaction(...)`, causing the second to crash with `ValueError: Staging transaction '…' is already active.` The retry path in `_process_chunk` (`packages/core/src/memex_core/services/ingestion.py:610`) appears to keep the same `chunk_txn_id` across attempts, which lets a parallel batch race onto the same id — or lets a retry re-enter staging that the previous attempt left partially open.

Version: `v0.1.11b`.

## Evidence

Observed during simultaneous batch jobs `bg-batch-f8893a1a2eee` and `bg-batch-ddf012c7ef26`:

```
[error] Failed to begin file staging: Staging transaction '5e209192-718c-a919-a2fd-5b39a9bf6a03' is already active..
  File "/app/packages/core/src/memex_core/services/ingestion.py", line 568, in ingest_batch_internal
    processed_ids = await self._process_chunk(
  File "/app/.venv/lib/python3.12/site-packages/stamina/_core.py", line 898, in async_inner
    async for attempt in retry_ctx.with_name(name, args, kw):
  File "/app/.venv/lib/python3.12/site-packages/tenacity/asyncio/__init__.py", line 166, in __anext__
    do = await self.iter(retry_state=self._retry_state)
  ...
  File "/app/packages/core/src/memex_core/services/ingestion.py", line 610, in _process_chunk
    async with AsyncTransaction(self.metastore, self.filestore, str(chunk_txn_id)) as txn:
  File "/app/packages/core/src/memex_core/storage/transaction.py", line 41, in __aenter__
    self.fs.begin_staging(self.txn_id)
  File "/app/packages/core/src/memex_core/storage/filestore.py", line 121, in begin_staging
    raise ValueError(f'Staging transaction {transaction_id!r} is already active.')
```

## Why this is worse than it looks

The failing chunk's exception was caught by the batch runner, which then logged:

```
[info] Batch job 2f1bd996-f450-4729-b94d-658c5f3db229 completed successfully.
```

That is **silent partial success**: the chunk was never ingested, but the batch is reported green. A loud failure would be preferable — as-is, data can be lost without any operator signal.

## Suspected mechanisms (please verify)

[...]

PR #43, opened 2026-04-24
fix(core): harden page_index extraction against wedges + silent batch failures

### LINKED ISSUE #40: Unbounded gap re-scan + missing concurrency cap wedges page_index extraction on small hosts
## Summary

On a single-worker memex server running on a Jetson Orin Nano (8 GB RAM, 4 GB GPU memory cap via `MEMEX_ONNX_GPU_MEM_LIMIT`), the page_index extractor hung indefinitely while processing a single 217,079-char / 58,311-token note. The event loop wedged for 2+ hours; every incoming HTTP request (including `/api/v1/health`) accepted TCP but never received a response. Nomad saw no crash and emitted no restart event; the Consul health check flipped to `critical: context deadline exceeded`.

Version: `v0.1.11b` (image `ghcr.io/jasperhg90/memex-jetson:0.1.11b`), `MEMEX_WORKERS=1`.

## Log evidence

```
[LLM Path] Scanning document (58311 tokens) in 3 chunks.
...
   > Omission Detected: 8984 chars. Re-scanning.
   > Omission Detected: 139209 chars. Re-scanning.
   > Omission Detected: 11149 chars. Re-scanning.
   > Omission Detected: 47619 chars. Re-scanning.
   > Tail Omission Detected: 6277 chars. Re-scanning.
```

The log stream went completely silent after the last re-scan was scheduled. No further output from the worker for the remainder of the observation window.

## Root cause

Three interacting issues in `packages/core/src/memex_core/memory/extraction/core.py`:

**1. Gap re-scan is unbounded.** At core.py:968 the LLM path calls:

```python
flat_headers = await self._detect_and_fill_gaps(flat_headers, full_text, max_gap_size=4000)
```

`_detect_and_fill_gaps` (core.py:1110) submits the *entire gap* as a single LLM scan call:

```python
if gap_size > max_gap_size:
    gap_text = full_text[current_idx : header.start_index]
    new_headers_tasks.append(
        self._process_single_chunk(gap_text, gap_context, offset=current_idx)
    )
```

No upper bound on gap size. In the run above, the 139,209-char gap went out as a single ~37K-token prompt.

**2. No concurrency cap.** Both `_scan_document_parallel` (core.py:1071) and `_detect_and_fill_gaps` (core.py:1146) use unbounded `asyncio.gather(*tasks)`. `ExtractionConfig.max_concurrency` (default 5) exists but only applies to fact extraction, not the page_index scan path.

**3. No LLM timeout.** The DSPy/LiteLLM call inside `_process_single_chunk` is not wrapped with a request timeout. If any one child of `asyncio.gather` hangs, the whole gather cannot return, and the single-worker event loop is wedged — blocking every unrelated request including the health check. This is what turns a slow extraction into an outage.

## Proposed fix (scoped to one PR)

1. Factor the "chunk a string into scan tasks" logic out of `_scan_document_parallel` into a helper such as `_build_scan_tasks(text, offset_base, max_scan_tokens, overlap_chars=200)`, and use it in **both** passes. Large gaps then fan out into N chunks of `scan_chunk_size_tokens` instead of a single monster call.

2. Add a new field on `PageIndexTextSplitting` (in `packages/common/src/memex_common/config.py`):

   ```python
   scan_max_concurrency: int = Field(
       default=5,
       description='Max concurrent LLM scan calls during page_index extraction.',
   )
   ```

   Wrap both `asyncio.gather` sites with a semaphore sized by this knob so small-GPU hosts can set it to 1 or 2. Intentionally **not** reusing `ExtractionConfig.max_concurrency` — that is a fact-extraction knob and conflating them will surprise users later.

3. Promote the hard-coded `max_gap_size=4000` at core.py:968 to a field on `PageIndexTextSplitting` such as `gap_rescan_threshold_tokens`, measured in tokens rather than characters so it tracks `scan_chunk_size_tokens` semantically.

4. Add a `request_timeout` to the `page_index_lm` (via DSPy/LiteLLM). A hung provider call should fail loudly, not wedge the worker. Without this, the first three fixes still leave the event loop exposed to a single bad request.

## Tests

- Feed `_detect_and_fill_gaps` a synthetic gap of `4 × scan_chunk_size_tokens` and assert it produces 4 scan tasks, not 1.
- Run a mocked scan with `scan_max_concurrency=1`, record enter/exit timestamps in the mock, assert tasks execute sequentially.

## Explicitly out of scope

- Auto-fallback to `SimpleTextSplitting` for oversized documents — that is a policy change and deserves its own issue.
- Smarter gap gating (regex-scan the gap for header-shaped lines before calling the LLM) — optimization, good follow-up.
- Raising `MEMEX_WORKERS` as a workaround — papers over the event-loop wedge rather than fixing it.

JasperHG90:
Fixes #40. Addresses the residual concern from #41 (the core collision was fixed by PR #39).

## Root causes (#40)

Three interacting issues in \`packages/core/src/memex_core/memory/extraction/core.py\` could wedge the event loop for hours on single-worker / small-host deployments:

1. **Unbounded gap re-scan.** \`_detect_and_fill_gaps\` submitted the entire gap (observed: 139K chars / ~35K tokens) as a single LLM call — one oversize prompt per gap.
2. **Unbounded concurrency.** Both \`_scan_document_parallel\` and \`_detect_and_fill_gaps\` called \`asyncio.gather(*tasks)\` without a cap. \`ExtractionConfig.max_concurrency\` exists but applies only to fact extraction.
3. **No LM request timeout on page_index.** \`ModelConfig.timeout\` (120s default, added in 8c4e100) was dropped at 4 of 5 \`dspy.LM(...)\` construction sites. A hung provider call wedged the event loop and blocked unrelated requests including \`/api/v1/health\`.

## Fix

- **Config (\`packages/common/src/memex_common/config.py\`)** — two new fields on \`PageIndexTextSplitting\`:
  - \`scan_max_concurrency: int = 5\` (ge=1). Separate from the fact-extraction knob on purpose; set to 1 on a Jetson Orin Nano.
  - \`gap_rescan_threshold_tokens: int = 2000\` (ge=500). Replaces the hardcoded \`max_gap_size=4000\` chars, measured in tokens to track \`scan_chunk_size_tokens\` semantically.

- **Extraction core (\`core.py\`)** — factored out \`_build_scan_tasks(text, max_scan_tokens, *, offset_base, context_prefix, overlap_chars)\` and apply it to **both** \`_scan_document_parallel\` and \`_detect_and_fill_gaps\`. Large gaps now fan out into N chunks instead of a single monster prompt. Introduced \`self._scan_semaphore\` acquired inside \`_process_single_chunk\` — mirrors the existing pattern at \`extraction/engine.py:211\` for fact extraction.

- **LM timeout** — fixed all 5 \`dspy.LM(...)\` sites to propagate \`timeout\` and \`num_retries\`:
  - \`memory/extraction/engine.py\` (\`_make_lm\`)
  - \`memory/reflect/reflection.py\` (×2)
  - \`memory/engine.py\` (×2)
  - \`api.py\` already did this — the rollout was incomplete.

- **Batch runner log (\`processing/batch.py:258\`, #41 residual)** — \`logger.info('Batch job X completed successfully.')\` was emitted even when \`failed_count > 0\`. Now conditionally emits \`logger.warning(...)\` with the first 3 errors when chunks failed. Job status contract unchanged.

## Test plan

- [x] \`just prek\` — ruff + mypy + format
- [x] \`uv run pytest packages/core/tests/unit/memory/extraction/test_page_index.py packages/core/tests/unit/test_batch_manager.py\` — 12/12 passing including 6 new regressions:
  - \`TestDetectAndFillGaps::test_oversized_gap_is_chunked_not_submitted_whole\`
  - \`TestDetectAndFillGaps::test_small_gap_below_threshold_is_skipped\`
  - \`TestDetectAndFillGaps::test_tail_gap_above_threshold_is_chunked\`
  - \`TestScanDocumentParallel::test_scan_respects_semaphore\`
  - \`TestMakeLmPropagatesTimeout::test_make_lm_passes_timeout_and_num_retries\`
  - \`test_run_job_partial_failure_logs_warning\`
- [x] \`uv run pytest packages/core/tests/unit/memory/ packages/core/tests/unit/storage/ packages/core/tests/unit/test_api_batch.py packages/core/tests/unit/test_api_ingest_path.py packages/core/tests/unit/test_batch_manager.py\` — 1032 passing, no regressions
- [ ] CI green on this PR

## Out of scope (flagged explicitly, not bundled silently)

- Auto-fallback to \`SimpleTextSplitting\` on oversized docs — policy change, deserves its own issue.
- Smarter gap gating (regex pre-scan for header-shaped lines) — optimization follow-up.
- New \`BatchJobStatus.PARTIAL\` — breaking change for downstream consumers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,323%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,221%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,317%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,323%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/packages/common/src/memex_common/config.py
+++ b/packages/common/src/memex_common/config.py
@@ -516,6 +516,22 @@ class PageIndexTextSplitting(BaseModel):
         default=None,
         description='Model for PageIndex LLM calls. If None, uses server default.',
     )
+    scan_max_concurrency: int = Field(
+        default=5,
+        ge=1,
+        description='Max concurrent LLM scan calls during page_index extraction. '
+        'Reduce on memory-constrained hosts (e.g. set to 1 on a Jetson Orin Nano) '
+        'to prevent GPU/RAM exhaustion. Separate from ExtractionConfig.max_concurrency, '
+        'which governs fact extraction.',
+    )
+    gap_rescan_threshold_tokens: int = Field(
+        default=2000,
+        ge=500,
+        description='Minimum gap size (in tokens) between detected headers that '
+        'triggers a secondary LLM re-scan to recover omitted headers. '
+        'Gaps larger than scan_chunk_size_tokens are sub-chunked using the same '
+        'chunking logic as the primary scan.',
+    )
 
 
 TextSplitting = Annotated[
--- a/packages/core/src/memex_core/memory/engine.py
+++ b/packages/core/src/memex_core/memory/engine.py
@@ -62,6 +62,8 @@ async def get_memory_engine(
             model=model_config.model,
             api_base=str(model_config.base_url) if model_config.base_url else None,
             api_key=model_config.api_key.get_secret_value() if model_config.api_key else None,
+            timeout=model_config.timeout,
+            num_retries=model_config.num_retries,
         )
         predictor = dspy.Predict(ExtractSemanticFacts)
         entity_resolver = EntityResolver()
@@ -109,6 +111,8 @@ def _build_contradiction_engine(config: MemexConfig) -> ContradictionEngine | No
             model=model_config.model,
             api_base=str(model_config.base_url) if model_config.base_url else None,
             api_key=model_config.api_key.get_secret_value() if model_config.api_key else None,
+            timeout=model_config.timeout,
+            num_retries=model_config.num_retries,
         )
         engine = ContradictionEngine(lm=lm, config=contradiction_config)
         logger.info(
--- a/packages/core/src/memex_core/memory/extraction/core.py
+++ b/packages/core/src/memex_core/memory/extraction/core.py
@@ -3,7 +3,7 @@ import hashlib
 import logging
 import re
 import asyncio
-from typing import cast
+from typing import Any, Coroutine, cast
 
 import dspy
 import regex as regex_lib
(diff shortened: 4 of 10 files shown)


JasperHG90:
Thanks for the review. Pushed 50836d1 addressing:

- **Finding 1 (bare `list` on `_build_scan_tasks`)** — applied. Return type is now `list[Coroutine[Any, Any, list[DetectedHeader]]]`.
- **Finding 3 (bare `list` on `new_headers_tasks`)** — applied. Same tightened annotation.
- **Finding 4 (`overlap_chars` never overridden)** — kept the parameter and expanded the docstring to explain when you'd override (dense gap rescans where the whole region is a single paragraph).
- **Finding 5 (test locality)** — moved `TestMakeLmPropagatesTimeout` from `test_page_index.py` to `test_extraction_engine.py` so it sits next to the function it tests.

**Pushed back on Finding 2 (eager `asyncio.Semaphore(N)` in `__init__`):** the claim that this emits a `DeprecationWarning` in Python 3.12 is not accurate — Python 3.10 removed the deprecated `loop=` parameter, but constructing `asyncio.Semaphore(value)` without a running loop is supported and warning-free in 3.12. No deprecation towards a 3.14 `RuntimeError` is in place for this pattern. `AsyncMarkdownPageIndex` is only instantiated from inside `index_document` (always under a running loop), so there's no realistic path where this matters. That said, the inline comment I'd written *was* describing CPython internals rather than intent — reworded it to say "the extractor is only ever instantiated from inside an async entry point, so constructing the Semaphore here is safe".

PR #48, opened 2026-04-24
fix(core): stop fabricating ghost IDs for virtual memory units (#47)

### LINKED ISSUE #47: Critical: Sync failure between Session Bootstrap and Live API (Ghost IDs)
### Bug Report: Memory-Graph Synchronization failure (Ghost/Phantom IDs)

**Description:**
A critical discrepancy has been identified between the pre-computed memory context (session bootstrap) and the live Memex API responses. The system is experiencing 'Ghost IDs' where the index returns memory units and note IDs that do not exist on the live server (404 Not Found), while the live API returns placeholder/dummy IDs for the same semantic facts.

**Observed Behavior:**
1. **Session Bootstrap**: The Assistant receives a `memory-context` containing a claim mapped to a specific `note_id` (e.g., `b3137f41-b282-4e7d-9e91-db49fdffd99a`).
2. **Live Recall**: Calling `memex_recall` for the same claim returns a different, dummy ID (e.g., `00000000-0000-0000-1cf2-e5666c664dbc`).
3. **Broken Lineage**: Attempting to use `memex_get_lineage` or `memex_read_note` with either the bootstrapped ID or the live-returned dummy ID results in a `404 Not Found`.
4. **Recovery**: The only way to find the actual data is via a fresh title/content search, which reveals the live note has a completely different ID (`b7ba79c8-25b5-7c4b-a80b-a78c367daa66`).

**Technical Impact:**
- **Lineage Failure**: Provenance chains are broken; the agent cannot trace facts back to their source documents.
- **Retrieval Gaps**: The agent may believe a fact is missing or corrupted because the associated IDs are invalid.
- **Hallucination Risk**: The agent may oscillate between contradictory IDs provided by the bootstrap vs. the live API.

**Suggested Investigation Path:**
- Audit the sync process between the `memory-context` generator (bootstrap phase) and the live MetaStore.
- Investigate why the API is returning `0000...` placeholder IDs instead of the actual canonical IDs.
- Check if a migration or re-indexing event caused a mismatch between the cached index and the physical storage.

**Evidence:**
- Recalled Obs ID: `00000000-0000-0000-1cf2-e5666c664dbc` (404)
- Bootstrapped Note ID: `b3137f41-b282-4e7d-9e91-db49fdffd99a` (404)
- Actual Live Note ID: `b7ba79c8-25b5-7c4b-a80b-a78c367daa66` (Active)

JasperHG90:
## Summary
- `_convert_mm_to_units` was constructing synthetic ids with `UUID(int=hash(str(model.id) + title) & (2**128-1))`. Python's `hash()` is 64-bit signed and salted per-process, so positive hashes produced `00000000-0000-0000-XXXX-XXXXXXXXXXXX` and negatives produced `ffffffff-ffff-ffff-XXXX-XXXXXXXXXXXX`, both unstable across restarts. It also set `note_id=model.id` — a MentalModel id in a Note FK slot, causing `memex_read_note` / `memex_get_lineage` to 404.
- Replace with `uuid5(_VIRTUAL_UNIT_NS, f'{model.id}:{title}')` (stable, full 128-bit), set `note_id=None`, and stash `model.id` as `unit_metadata['mental_model_id']`.
- Recall serializers (`_serialize_memory_unit`, `_serialize_memory_unit_full`, `_build_memory_unit_model`) now surface `virtual=true` + `mental_model_id` + `evidence_ids` on synthetic units so agents don't try to look them up and can reach the backing data via real evidence memory ids.
- `McpMemoryUnitBase` gains three optional, defaulted fields — backwards-compatible.

Closes #47.

## Test plan
- [x] Regression test asserts `id.int >> 64 != 0`, `note_id is None`, stability across calls, and `virtual`/`mental_model_id`/`evidence_ids` in both Hermes-plugin dict serializers and the MCP `_build_memory_unit_model`.
- [x] `just prek` (ruff + ruff-format + mypy) clean.
- [x] core unit (1858), common (109), mcp non-integration (309 + 2 new), hermes non-integration (272 + 4 new), root E2E (35), core integration (283) — all green.
- [ ] hermes_integration suite (requires `hermes-agent` git dep) — skipped in this env.
- [ ] Live agent exercise against a running Memex server with a MentalModel-surfacing query.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,221%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,506%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/packages/core/src/memex_core/memory/retrieval/engine.py
+++ b/packages/core/src/memex_core/memory/retrieval/engine.py
@@ -63,8 +63,9 @@ K_RRF = 60
 CANDIDATE_POOL_SIZE = 60
 
 # Namespace for deterministic virtual MemoryUnit ids synthesized from
-# MentalModel observations. Derived from uuid5(NAMESPACE_URL, ...) — stable
-# across processes, unlike Python's PYTHONHASHSEED-salted hash().
+# MentalModel observations. Stable across processes, unlike Python's
+# PYTHONHASHSEED-salted hash(). Reproducible via:
+#   uuid5(uuid.NAMESPACE_URL, 'memex/retrieval/virtual-mental-model-observation')
 _VIRTUAL_UNIT_NS = UUID('bf63d7e5-1e6a-5cf4-9f33-2e85d3a48d38')
 
 
--- a/packages/core/src/memex_core/memory/retrieval/engine.py
+++ b/packages/core/src/memex_core/memory/retrieval/engine.py
@@ -3,7 +3,7 @@ import logging
 import time
 from collections import defaultdict
 from dataclasses import dataclass, field
-from uuid import UUID
+from uuid import UUID, uuid5
 import asyncio
 from datetime import datetime, timezone
 from typing import Any, Sequence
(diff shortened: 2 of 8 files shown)


JasperHG90:
Addressed the Hermes review in 594c870:

- **MEDIUM — `mental_model_id` unguarded UUID coercion** (server.py): guarded with `try/except (ValueError, TypeError)` mirroring the `evidence_ids` pattern. Corrupted metadata now yields `mental_model_id=None` rather than raising.
- **MEDIUM — `evidence_ids` type inconsistency**: added an explanatory comment at the MCP coercion boundary. The design is intentional — metadata stores strings for JSONB round-trip stability; MCP coerces to `UUID` at the typed response edge. A full-pipeline type refactor is out of scope for this fix.
- **LOW — `_VIRTUAL_UNIT_NS` provenance comment**: updated to include the exact seed (`uuid5(NAMESPACE_URL, 'memex/retrieval/virtual-mental-model-observation')`) so the constant is reproducible.
- **LOW — `_full` test evidence_ids count**: bumped to two ids to pin list-length integrity, matching the short-serializer test.

**Rejecting HIGH — `metadata` may be undefined in `_serialize_memory_unit_full`** as a false positive. `metadata` is assigned on line 2111 of the function (7 lines above where `is_virtual` reads it), defined by the pre-existing `metadata = getattr(unit, 'metadata', None) or {}`. The reviewer saw only the diff hunk and didn't have the surrounding function context. No runtime `NameError` is possible.

PR #52, opened 2026-04-25
fix(core): cap extraction concurrency + batch dedup + wedge watchdog

### LINKED ISSUE #50: page_index Fast Path still wedges on Jetson: unbounded asyncio.gather in 3 sites not fixed by #43
## Summary

`page_index` extraction wedged the event loop again on `memex-jetson:0.2.0` (which **does** include #43). The wedge happened in the **Fast Path** (`_refine_tree_recursively` → recursive Deep Dive), which #43 did not touch — #43 only hardened the Slow Path (`_scan_document_parallel` / `_detect_and_fill_gaps`).

Three unbounded `asyncio.gather(*tasks)` call sites remain in `packages/core/src/memex_core/memory/extraction/core.py`. Each fans out a Gemini call per node/leaf/block with no semaphore. On a small upstream connection budget (or any rate-limit / timeout situation) this still hangs the whole worker indefinitely.

## Reproduction (production log excerpt)

Single-worker memex on Jetson Orin Nano, doc = 197,569 chars, 15 regex-detected headers, 92% coverage → Fast Path:

```
02:13:54.836 [Fast Path] Recursive refinement
02:13:54.837    > Deep Dive: Refining '4. Role in the Extraction Pipeline' (41711 chars)
02:13:54.845    > Deep Dive: Refining 'Page Index Inspection' (67450 chars)
02:13:54.853    > Deep Dive: Refining 'Summary for Trading Application' (63390 chars)
02:13:57.606    > Deep Dive: Refining 'References' (41429 chars)
02:13:57.779    > Deep Dive: Refining 'References' (62998 chars)
02:13:57.840    > Deep Dive: Refining 'Towards a Science of Scaling Language Model Agent Systems' (35100 chars)
```

Six refinement tasks (each = one Gemini call wrapping a 35–67 K-char chunk) entered flight in ~3 seconds. Worker became unresponsive; required manual restart 4 h 45 m later. On reboot, `reconcile_interrupted_jobs` flagged two interrupted batches (`d2e0066b…`, `82bbf403…`) which had been processing the *same* note seconds apart.

## Unbounded gather sites in current `main`

| Method | Line | Notes |
|---|---|---|
| `_refine_tree_recursively` | ~1191 | Fans out `_process_single_node_refinement` for every node; recursive — children also fan out unbounded. |
| `_generate_summaries_parallel` | ~1263, ~1266 | Two unbounded gathers: leaf summaries, then parent summaries. |
| `_generate_block_summaries` | ~1329 | Block-summary tasks gathered without a cap. |

`__init__` exposes `scan_max_concurrency` and `gap_rescan_threshold_tokens` (good), but `self._scan_semaphore` is only consumed by `_process_single_chunk`. Refinement / summary / block paths bypass it entirely.

## Compounding factor: GPU memory exhaustion 6 minutes earlier

At 02:07:56 — six minutes before the wedge — the survey reranker (`services/search.py:270`, `asyncio.to_thread(self.reranker.score, query, formatted_texts)`) hit ~25 consecutive `NvMapMemAllocInternalTagged: 1075072515 error 12` (≈1 GB allocations failing, ENOMEM), culminating in `CUDNN_STATUS_INTERNAL_ERROR` on `cudnnCreate(&cudnn_handle_)`. The HTTP request returned 500 cleanly, but the cuDNN context for that worker is now in an unknown state.

This is on a Jetson Orin Nano (8 GB **unified** CPU+GPU memory) with:

- cgroup `memory_max = 7000` MiB
- `MEMEX_ONNX_GPU_MEM_LIMIT = 4 GiB` (counts against the same pool)
- `RERANKER_BATCH_SIZE = 8`, `EMBEDDING_BATCH_SIZE = 64`

Once the 4 GiB arena is committed, the residual ~3 GiB has to cover Python heap + cuDNN workspaces (which want ~1 GiB chunks). That's the immediate cause of the NvMap failures. It's plausible but unproven that a stuck `asyncio.to_thread` in this degraded state contributed to the later loop wedge — the page_index calls themselves are pure Gemini HTTP, but other code paths in the same loop still touch ONNX.

## Proposed fix (primary)

Apply the same semaphore pattern #43 used for `_process_single_chunk` to **every** fan-out in `core.py`:

1. Reuse `self._scan_semaphore` (or add a `self._refine_semaphore` / `self._summary_semaphore` if you want them tunable separately) and wrap each task body in `async with sem:` before the LLM call.
2. Bound `_refine_tree_recursively` *and* the recursive path inside `_process_single_node_refinement` so deeply-nested trees can't spawn N×M concurrent refinements.
3. Promote `max_node_length` (currently 5000 chars, default refine threshold) to a config field, and split sections > some token cap into multiple chunked refinement calls — same idea as `gap_rescan_threshold_tokens`.
4. Audit other extraction sub-modules for the same pattern (`_generate_*` in `core.py`, anything in `extraction/pipeline/`).

## Proposed fix (secondary, optional)

- **Enqueue dedup:** `BatchManager.create_job` does not check whether any incoming `note_id` is already in a `PENDING` / `RUNNING` job. Two batches processing the same 197 K-char doc 4 s apart is wasted work and doubles the wedge probability. Suggest a unique-on-active-note-id check or merge-into-existing-job behaviour.
- **Reconciliation:** `reconcile_interrupted_jobs` marks interrupted jobs `FAILED` with "Please resubmit" — fine, but combined with no enqueue dedup it lets the operator unintentionally double-submit while the original is still wedged.
- **Sizing guidance for unified-memory hosts:** document recommended (`ONNX_GPU_MEM_LIMIT`, `RERANKER_BATCH_SIZE`, `EMBEDDING_BATCH_SIZE`, container `memory_max`) tuples for Jetson Orin Nano / Nano Super. Current defaults assume discrete VRAM. On 8 GB unified memory, 4 GiB arena + batch=8 reranker is very close to the edge under any concurrent extraction load.

## Acceptance

- [ ] Concurrency cap enforced at every LLM fan-out in `extraction/core.py` (regression test that asserts ≤ N in-flight calls under a synthetic large-doc input).
- [ ] Repro doc (197 K chars, 15 headers, 6 sections > 30 K chars) completes in bounded time on a single worker, no wedge.
- [ ] No regression on Slow Path behaviour from #43.
- [ ] Optional: dedup or warn at `BatchManager.create_job` when overlapping notes are submitted.

## Environment

- `ghcr.io/jasperhg90/memex-jetson:0.2.0` (release v0.2.0, commit `c21f192`, contains #43 / `3951edf`)
- Nomad alloc on jetson-orin-nano, single worker (`MEMEX_WORKERS=1`)
- Default `gemini/gemini-3.1-flash-lite-preview`
- `scan_max_concurrency` left at default 5

cc #40 #43

### LINKED ISSUE #51: RFC: prevent extraction wedges via concurrency caps + cancellable LLM ops
## Status

Draft RFC. Replaces the narrower bug filed in #50.

## Context

We have now had two consecutive single-worker wedges where `extraction.page_index` stops responding for hours instead of completing or failing fast. Symptom is the same; the *cause* differs each time, so an incremental fix-as-we-go approach has not converged.

- #40 / #43 — fixed unbounded `asyncio.gather` + missing concurrency cap on the **Slow Path** (`_scan_document_parallel` / `_detect_and_fill_gaps`) and added an LM `timeout` to several call sites.
- #50 — same wedge pattern, but now in the **Fast Path** (`_refine_tree_recursively` and the two unbounded gathers in `_generate_summaries_parallel` / `_generate_block_summaries`). PR #43 did not touch these.

The deeper issue is that the codebase has the same risky pattern in many places, and our defenses (LM `timeout`, the 600 s outer `asyncio.wait_for` in `index_document`) are not actually sufficient on their own. We should fix this once, structurally, and add the diagnostics needed to catch regressions cheaply.

## What we know vs. what we hypothesize

**Confirmed from the code (current `main`):**

1. `index_document` *does* wrap `aforward` in `asyncio.wait_for(..., timeout=600)` (extraction/core.py:1051). So a single hung call should bail in ≤ 10 min.
2. `run_dspy_operation` (memex_core/llm.py) executes the predictor either via `predictor.acall(...)` if available, otherwise via `await asyncio.to_thread(predictor, **kwargs)`. There is no per-call timeout inside `run_dspy_operation` itself.
3. The semaphore added in #43 (`self._scan_semaphore`) is acquired **only** inside `_process_single_chunk`. Refinement, leaf summary, parent summary, and block summary fan-outs do not touch it.
4. `dspy.ChainOfThought` predictors are sync in DSPy ≥ 2.x; whether they expose `acall` depends on the DSPy version. If they do not, every LLM call in extraction goes through `asyncio.to_thread`.

**Hypothesis for the 4 h 45 m wedge in #50 (NOT proven from logs):**

The 600 s outer timeout fires, but `asyncio.wait_for` only cancels the *coroutine*; the threads dispatched by `asyncio.to_thread` keep running until their socket-level operation completes (LiteLLM/httpx may or may not have a low socket timeout, depending on how the LM was constructed). Each wedged batch pins ~6 threads in the default executor (`ThreadPoolExecutor(max_workers=min(32, os.cpu_count()+4))` ⇒ 10 on Jetson Orin Nano). With no enqueue dedup, repeated automatic/manual retries pile more stuck threads on, until the executor is saturated and *every* sync-offloaded path (health checks, NER, embeddings via `to_thread`) blocks.

This is the part I can't prove from the logs we have. The RFC's diagnostic proposals below are designed to make this verifiable next time.

## Goals

1. No fan-out site in `extraction/` can issue more than `N` concurrent LLM calls, where `N` is configurable per stage.
2. A wedged document must surface as a fast, loud failure — minutes, not hours — even when DSPy is sync-offloaded.
3. When a wedge does happen, we capture a stack dump + in-flight task counts so we can root-cause without having to repro.
4. Memex remains generic — Jetson is one deployment among many. All caps and timeouts are configurable, with sensible defaults that match today's behaviour for hosts that don't tune them.

## Non-goals

- Auto-fallback strategies (bypass to Slow Path when Fast Path takes too long, etc.). That's a separate design.
- Rewriting DSPy / LiteLLM internals.
- Deciding the right defaults for small hosts. That's a deployment concern; this RFC only ensures the levers exist.

### 1. Cap concurrency at every fan-out in `extraction/core.py`

Add three optional, defaulted ctor params to `AsyncMarkdownPageIndex`:

```python
def __init__(
    self,
    lm: dspy.LM,
    *,
    scan_max_concurrency: int = 5,
    refine_max_concurrency: int = 5,
    summarize_max_concurrency: int = 5,
    gap_rescan_threshold_tokens: int = 2000,
    refine_chunk_threshold_tokens: int | None = None,  # see §2
) -> None:
```

Construct one semaphore per stage and gate every existing fan-out:

- `_refine_tree_recursively` → `_process_single_node_refinement` wraps its LLM body in `async with self._refine_semaphore`.
- `_generate_summaries_parallel`'s leaf and parent gathers wrap each task body in `async with self._summary_semaphore`.
- `_generate_block_summaries` wraps each task body in the same summary semaphore (or a dedicated `_block_summary_semaphore` if we want it tunable separately — I lean toward shared to keep the surface small).

`run_dspy_operation` already accepts an optional `semaphore=` argument; we can plumb through that path rather than nest `async with`. Either is fine; the test should assert `≤ N` in flight, not the implementation detail.

For consistency, also propagate these from `index_document(...)`'s signature so callers in `extraction/engine.py` can configure them via `ExtractionConfig`.

### 2. Bound recursive refinement size

Today, `_process_single_node_refinement` re-submits the entire section text (up to `max_node_length` worth of recursion) as a single LLM call when `node_len > max_len`. A 67 K-char section becomes a 16 K-token prompt in one shot. Same failure mode as the gap re-scan that #43 chunked.

Add `refine_chunk_threshold_tokens` (default `None` → existing behaviour preserved). When set, sections above that threshold are split into overlapping chunks and refined in parallel under the semaphore, then merged the same way `_detect_and_fill_gaps` merges scan results.

### 3. Make LLM ops actually cancellable (or be honest that they aren't)

Two parts:

a. **Push timeouts to the network layer.** `dspy.LM` accepts `timeout` (PR #43 added it to several construction sites). Verify that this actually maps to a LiteLLM/httpx **socket** timeout, not just a wall-clock guard around the high-level call. If it doesn't, set a `request_timeout` on the LM's underlying client explicitly.

b. **Wrap `asyncio.to_thread` paths in a per-op `asyncio.wait_for`.** The outer 600 s budget at the pipeline level is too coarse — by the time it fires, the thread pool may already be saturated. Add a per-LLM-call timeout inside `run_dspy_operation` (configurable, default e.g. 180 s). When it fires, raise immediately even if the thread is still running. Document explicitly that the underlying thread will continue to drain the socket; this is acceptable as long as the network-layer timeout in (a) eventually closes it.

### 4. Diagnostics

a. **Per-stage in-flight gauge.** Increment on enter / decrement on exit for each semaphore-gated section. Emit as a Prometheus gauge (`extraction_inflight{stage="refine"}`) and log at info every N seconds while > 0.

b. **Watchdog stack dump.** Register a background task that, if no extraction event has been logged for > M minutes (configurable, off by default), calls `faulthandler.dump_traceback(file=...)` to capture *all* thread stacks. Emits one log line + one trace, then re-arms. This is the single highest-leverage thing we can add to avoid another hour-long forensic guessing game.

c. **Stage timing.** Log structured `stage_complete` with `stage`, `session_id`, `duration_ms`, `tasks_in_flight_max` for each fan-out. Five extra log lines per document; trivial overhead.

### 5. Prevent overlapping batches on the same notes (optional follow-up)

`BatchManager.create_job` does not check whether any incoming `note_id` is currently being processed. The wedge in #50 ran two concurrent batches on the same 197 K-char document. Either reject overlapping submissions with a 409, or coalesce them into the existing job. Either way, document the chosen semantics.

This is independent of the rest of the RFC and could ship separately.

## Configuration surface (proposed `ExtractionConfig` additions)

| Field | Default | Notes |
|---|---|---|
| `scan_max_concurrency` | 5 (existing) | Slow Path scanner |
| `refine_max_concurrency` | 5 | Recursive refinement |
| `summarize_max_concurrency` | 5 | Leaf + parent summaries; reused by block summaries |
| `refine_chunk_threshold_tokens` | `None` | Off by default; opt-in to chunking large sections |
| `per_call_timeout_seconds` | 180 | Hard timeout inside `run_dspy_operation` |
| `pipeline_timeout_seconds` | 600 (existing constant; promote to config) | Outer `asyncio.wait_for` budget |
| `wedge_watchdog_seconds` | `None` | When set, dump tracebacks if no extraction event in this window |

All defaults preserve current behaviour for current operators (with the *exception* of `per_call_timeout_seconds`, which is strictly tighter than today's "no per-call timeout" — discuss).

## Acceptance / validation

- Unit test: synthetic doc with 20 oversized sections runs through a stubbed LM and asserts that no more than `refine_max_concurrency` calls are ever in flight (use a counter LM).
- Unit test: stub LM that hangs forever — pipeline returns within `per_call_timeout_seconds × 1.5` instead of `pipeline_timeout_seconds`.
- Integration test: 197 K-char doc with 15 headers and 6 large sections completes on a single worker without hanging.
- No regression on Slow Path tests added in #43.

## Open questions

1. Should the per-stage caps share one semaphore, or three? Three is more flexible; one is simpler. Vote: three, defaulted equal.
2. Does `dspy.LM(timeout=...)` actually plumb through to `httpx`'s socket timeout under LiteLLM? Needs to be checked before §3a is closed.
3. For (5), is "reject overlap" or "coalesce" correct? Probably reject + return existing job_id, but worth picking deliberately.

## Migration

Pure additions to constructor / config. Existing callers keep working unchanged. The one behavioural change is the default `per_call_timeout_seconds=180` — operators who deliberately rely on long-running LLM calls would need to bump it. We can ship that off by default (`None`) in v1 and turn it on in a later minor.

## Refs

- #40 — initial wedge report (Slow Path)
- #43 — fix for #40 (Slow Path only)
- #50 — recurrence (Fast Path)

JasperHG90:
## Summary

- **Closes #50, addresses #51 §1–4 (wedge fix).** Caps the four unbounded `asyncio.gather` fan-outs in `extraction/core.py` (refine + 3 summary stages) behind per-stage semaphores; plumbs `ModelConfig.timeout` into every `dspy.LM(...)` construction so httpx has a real socket deadline; adds an opt-in OS-thread watchdog that calls `faulthandler.dump_traceback` when no gated stage decrements for `wedge_watchdog_seconds`.
- **Closes #51 §5 (batch dedup).** New `BatchJob.input_note_keys` JSONB column (Alembic 021, GIN-indexed, PG ≥ 11 guard); `JobManager.create_job` derives keys from incoming DTOs, takes a `pg_advisory_xact_lock` per vault, queries for overlap with `PENDING|PROCESSING` jobs, and either inserts or raises `OverlapError`. The batch endpoint translates that to **HTTP 409** + `Location: /api/v1/ingestions/{job_id}`. The error carries `overlapping_keys` — the actual subset of incoming keys that hit the in-flight job.
- **Sync-offload hardening.** New `_offload` module with per-model semaphores (NER / embeddings / reranker) sized from `ServerConfig`; every `asyncio.to_thread(...)` to a model wrapped in `wait_for(...)`; warmup ordering at startup so cold-cache calls don't race the cap.
- **Observability.** `memex_extraction_inflight` and `memex_sync_offload_inflight` Prometheus gauges, labelled by stage.
- **Docs.** `docs/how-to/memory-budget.md` — sizing guidance for memory-constrained hosts.

## ⚠️ Behavioural changes operators must know about

1. **Concurrency-cap defaults raised for capable hosts.**
   - `PageIndexTextSplitting.scan/refine/summarize_max_concurrency`: **5 → 20** each.
   - `ServerConfig.reranker/embedding/ner_max_concurrency`: **4 → 16** each, and the `le=4` validation cap is **dropped**.
   - **Constrained hosts (Jetson, Pi, low-RAM containers) must set lower values explicitly** — see `docs/how-to/memory-budget.md`. Without an override, a memory-constrained deployment that previously ran fine on default-5/default-4 will now fan out 4× and may OOM.
2. **Duplicate batch submissions return 409 + `Location`** instead of silently starting a parallel job. Clients with retry-on-error logic must treat 409 (with `Location: /api/v1/ingestions/{existing_job_id}`) as "already in flight; poll the location".
3. **Migration `021_batch_jobs_input_note_keys` must run before the dedup query works** (`just db-upgrade`). Online migration on PG ≥ 11; aborts loudly on PG < 11.

## Test plan

- [ ] CI green (unit + integration suites)
- [ ] `uv run pytest packages/core/tests/unit -q` locally
- [ ] Repro doc (197 K-char, 6 large sections) on a single-worker memex against a real Gemini endpoint — confirm bounded completion
- [ ] Submit identical batch twice to `POST /api/v1/ingestions/batch` — second call returns 409 with `Location` header
- [ ] Set `extraction.wedge_watchdog_seconds: 60` and stub a slow LM — confirm `faulthandler` dump
- [ ] `/metrics` shows `memex_extraction_inflight{stage="..."}` and `memex_sync_offload_inflight{stage="..."}` gauges

## Notes for reviewer

- **CI gap surfaced during review:** the `python-tests` workflow only runs `tests/`, not `packages/*/tests/`. Worth filing a follow-up — Hermes caught a default-value assertion mismatch in `packages/common/tests/test_config_structure.py` that CI never would have flagged.
- Follow-up deferred: umbrella `extraction_max_concurrency` knob that defaults the three per-stage caps from one field.


JasperHG90:
Thanks for the review. Addressed in `c384784`:

**Critical** ✅ — `test_config_structure.py` assertion bumped to `== 20`. Caught a real gap: the CI `python-tests` job only runs `tests/`, never `packages/*/tests/`, so this would have stayed broken.

**High (semaphore kwarg)** — Pushing back. `run_dspy_operation` already accepts `semaphore: asyncio.Semaphore | None = None` (`packages/core/src/memex_core/llm.py:57` on main pre-PR). The diff doesn't show the parameter because it wasn't added by this PR — only new call sites that pass it. Not a bug.

**High (scan vs refine `async with` style)** — Acknowledged. Not changed in this commit because tightening the scan site to a combined `async with self._scan_semaphore, _instrument('scan'):` would also shrink the semaphore-held region (currently covers post-LLM header processing too), which is a behavioural change rather than a pure style normalization. Worth a follow-up; not load-bearing for this PR.

**Medium** ✅
- `BatchJob.input_note_keys`: `default=[]` → `default_factory=list`.
- `_summarize_single_node` exception handler — kept narrow. `_instrument._read_gauge_value` already catches `Exception` internally and returns 0; the only path that could surface `_instrument`-internal errors is already squelched, so the narrow handler in `_summarize_single_node` doesn't risk swallowing.
- `test_int_alembic_021.py` f-string DDL: comment added explaining `db_name` is `'mig021_' + secrets.token_hex(6)` (fixed prefix + hex chars).

**Low** ✅
- Watchdog TOCTOU: comment added on `check_once` noting the gauge read happens outside the lock and a one-shot false-positive on a just-finished stage is acceptable.
- `hashtext` collision: `TODO: hashtextextended` marker added.
- README badge: was an environment artifact — `memex_hermes_plugin` wasn't installed in the venv that ran the local pre-commit hook, so its tests were silently dropped from the count. Restored to the post-`uv sync --all-packages` figure.
- Dep pin tightening: no action needed (your read agrees with the rationale).

Re-review when you have a moment — `/review` if you want Hermes to take another pass.

JasperHG90:
Round-2 review addressed in `5308e83`. PR body now flags the behavioural changes (cap defaults, 409 dedup, migration prereq).

**Real findings — fixed:**

- **Medium (litellm catch ordering)** — comment added at the `except TimeoutError` block in `llm.py` documenting the dependency on `litellm.exceptions.Timeout` not being a `TimeoutError` subclass; if a future litellm/openai-sdk release restructures that hierarchy, both blocks must be revisited.
- **Medium (dep upper bounds)** — inline rationale comments on `dspy<3.2.0`, `httpx<0.29.0`, `litellm<1.81.0` in `packages/core/pyproject.toml`.
- **Medium (`overlapping_keys` always empty)** — the overlap query now also selects the existing job's `input_note_keys`; `OverlapError` carries `sorted(set(input_keys) & set(existing_keys))`. Operators receiving 409 see exactly which incoming notes hit the in-flight job. New positive test `test_create_job_overlapping_keys_subset_returned`.
- **Low (`_instrument` docstring wording)** — re-worded to "_instrument is entered after the semaphore is acquired", with both `async with A: B:` and `async with A, B:` shown as equivalent forms.
- **High (default change is a behavioural break)** — flagged in the PR body now under "Behavioural changes operators must know about". Defaults stay at 20 / 16 (deliberate user-side decision: optimise for the common case, document the override path), but the migration guidance for memory-constrained hosts is now explicit.

**False positives — pushing back again:**

- **Critical: "`semaphore=` kwarg passed to `run_dspy_operation` but not in its signature"**
- **High: "Summarize fan-outs ungated"**

Both stem from the same misread. `run_dspy_operation` *does* accept `semaphore` and *does* acquire it — has done since pre-PR. Quoting the actual source:

```python
# packages/core/src/memex_core/llm.py:53-60
async def run_dspy_operation(
    lm: dspy.LM,
    predictor: dspy.Module,
    input_kwargs: dict[str, Any],
    semaphore: asyncio.Semaphore | None = None,    # <-- the parameter
    operation_name: str = 'dspy',
    timeout: int = 180,
) -> Any:
```

```python
# packages/core/src/memex_core/llm.py:113-118
try:
    if semaphore:
        async with semaphore:                       # <-- the gating
            result = await asyncio.wait_for(_execute(), timeout=timeout)
    else:
        result = await asyncio.wait_for(_execute(), timeout=timeout)
```

Neither change is in this PR's diff because both lines are in `main` pre-PR. The summarize/block_summarize fan-outs that pass `semaphore=self._summary_semaphore` *are* gated end-to-end. Confirmed by the existing `test_leaf_summary_observes_summarize_max_concurrency`, `test_parent_summary_observes_summarize_max_concurrency`, and `test_block_summary_observes_summarize_max_concurrency` tests in `packages/core/tests/unit/memory/extraction/test_page_index.py`, which assert the cap is enforced via this exact path.

If round-3 raises this again, please flag the specific contradicting evidence — would help to know whether the patch-only context is a fundamental review-tool limitation or something I can mitigate by, e.g., re-stating the existing signature in a comment at each call site.

**Lows skipped:**

- `_summarize_single_node` exception narrowness — `_instrument._read_gauge_value` already catches broad `Exception` internally and returns 0, so the narrow handler in the caller can't swallow `_instrument`-internal errors. Documented in `_read_gauge_value`'s docstring.
- `h.id = idx` redundancy in `_process_single_node_refinement` — pre-existing code outside this PR's scope.
- `_read_gauge_value` traceback noise — gauge reads are expected to never fail; a transient failure is rare enough that a full traceback is the right log shape.

JasperHG90:
Round-3 addressed in `70f93ca`. Hermes caught a real bug this round — credit where due.

**High #1 was correct (gauge ordering on summarize sites)** ✅
The three summarize methods (`_summarize_single_node`, `_summarize_parent_node`, `_summarize_single_block`) entered `_instrument('summarize')` *before* the semaphore was acquired (the semaphore was being acquired inside `run_dspy_operation` via the kwarg). With cap=2 and 8 pending tasks, the gauge would briefly read 8 — counting blocked-on-cap tasks as "in-flight". Fixed by rewriting all three sites to match the scan/refine pattern:

```python
async with self._summary_semaphore, _instrument('summarize'):
    pred = await run_dspy_operation(
        lm=self.lm,
        predictor=self.summarizer,
        input_kwargs={...},
        operation_name='extraction.summarize',
    )
```

Side effect: dropped the `semaphore=` kwarg from these three call sites (gating now happens at the outer `async with`), which incidentally clears the false-positive Critical/High you've been flagging across the last three rounds. The kwarg + internal acquire still exists in `run_dspy_operation` for other callers; just unused here now.

**Medium #1 (constructor defaults)** ✅ — `AsyncMarkdownPageIndex.__init__` and `index_document(...)` now default `scan/refine/summarize_max_concurrency=20`, matching `PageIndexTextSplitting`. Direct instantiation no longer silently downgrades to the old cap.

**Medium #2 (`get_circuit_breaker` comment)** ✅ — comment rephrased: in-module callers (the exception handlers in `run_dspy_operation`) read the module global directly because the rebind happens at module level; external importers should use `get_circuit_breaker()` to avoid binding a stale reference at import time. Both patterns are correct in their respective contexts.

**Low #1 (`watchdog._dump_path`)** ✅ — added public `dump_path` property; `server/__init__.py` uses it.

**Skipped:**

- **Medium #3 (`_offload` thread-safety)** — `configure_offload_semaphores` runs once at server startup; runtime reconfiguration isn't a real path. The four assignments are sequential without a lock, but a partially-reconfigured read would only matter if a getter ran concurrently with reconfigure, which doesn't happen.
- **Low #2 (audit log non-zero overlap test)** — fair point but out of scope; the new `test_create_job_overlapping_keys_subset_returned` covers the subset-computation contract end-to-end at the unit level.
- **Low #3 (alembic test f-string guard)** — the existing comment already documents that `db_name = 'mig021_' + secrets.token_hex(6)` (fixed prefix + hex chars only); a runtime assert would be belt-and-suspenders.

If round-4 still flags `semaphore=`, please look at `packages/core/src/memex_core/llm.py:53-60` (signature) and `:113-118` (acquire) directly — those lines are unchanged across this entire PR.

JasperHG90:
Round-4 addressed in `467ae7d`. Both Mediums fixed; Lows mostly skipped with rationale.

**Medium #1 (litellm catch-ordering self-check)** ✅ — added an `assert not isinstance(e, TimeoutError)` in the litellm branch. Zero runtime cost on the happy path; fires loudly the moment the assumption is violated.

**Medium #2 (`ValueError` from empty `notes` returning 500)** ✅ — `ingest_batch` now catches `ValueError` separately from the broad `MemexError` catch and surfaces 422. Pydantic still rejects at the request layer; this handles internal callers.

**Lows skipped:**

- **Low #1 (`_process_single_chunk` semaphore+`_instrument` pattern):** intentionally not unified. The scan site's `_instrument('scan')` is *deliberately* narrower than the semaphore so the post-LLM header-validation loop (text-index lookups, fuzzy regex fallback) isn't counted as in-flight in the gauge. The other four sites don't have this distinction — they only do the LLM call inside the gated region. A combined `async with self._scan_semaphore, _instrument('scan'):` would extend the gauge to cover the validation loop, which is a behavioural change rather than style normalisation.
- **Low #2 (`_offload` getter `assert`):** `_require_configured()` checks `_CFG`, but `_CFG` is set in the same statement that sets the semaphores in `configure_offload_semaphores`, so `_CFG is not None` is a sufficient invariant for the semaphore. The `assert` exists for mypy's type-narrowing — it's stripped under `-O` in production but not load-bearing. Could replace with explicit raise, but the current shape isn't actively wrong.
- **Low #3 (watchdog construction-time-firing operator note):** the watchdog is opt-in (default `wedge_watchdog_seconds: None`); operators who set it explicitly are choosing observability over false-quiet. The construction-time fallback is documented in the class docstring and is the conservative correct behaviour. Adding a note in `memory-budget.md` would be reasonable but isn't load-bearing.
- **Low #4 (autouse fixtures mutating module globals):** the comment is informational. Tests run sequentially in this project; `pytest-xdist` workers don't share globals. If the project adopts in-process parallelism later, this fixture (and the prek-cache cross-pollution we already documented in INCIDENT-2) would need revisiting.
- **Low #5 (`_read_gauge_value` exception masking):** existing code already does `logger.exception(...)` (ERROR level + traceback). Hermes's suggestion to also log at debug is actually a downgrade — the current shape is louder, not quieter.
- **Dependency tracking issues:** good idea operationally but out of scope for the PR. The pin comments name what to revalidate; happy to file follow-up issues separately.

Per Hermes's own summary: "0 Critical, 0 High; the concurrency model is sound; the test coverage is extensive; the documentation is excellent." Considering this PR ready for human review/merge.

PR #60, opened 2026-04-26
fix(core): reject empty path in /resources and filestore reads

JasperHG90:
## Summary

- Production was returning 500s from `GET /api/v1/resources/` because an empty `{path:path}` propagated through the service chain into `s3fs._cat_file`, which boto then rejected with `ParamValidationError: Invalid length for parameter Key, value: 0, valid min length: 1`.
- Root cause: `validate_path_safe` returns the storage root for an empty key (intentional — `check_connection` lists the bucket root), but `BaseAsyncFileStore.load` and the `/resources` route never rejected empty inputs from public callers.
- Fix is a defence-in-depth pair: `load` now raises `FileNotFoundError` on empty/whitespace/all-slash keys (the route already converts that to a clean 404); `exists` / `is_dir` return `False` for the same; the route shortcuts an obvious empty path to 404 before entering the service layer. `validate_path_safe` / `join_path` are unchanged.

## Test plan

- [x] `packages/core/tests/unit/storage/test_filestore.py` — new parametrized tests assert `load`/`exists`/`is_dir` reject `''`, `'/'`, `'///'`, `'   '`, plus a regression test that `check_connection` still works (it bypasses the guarded methods).
- [x] `tests/test_e2e_resources.py` — new parametrized integration test asserts `GET /api/v1/resources/{empty}` returns **404** with body `{"detail": "...empty..."}` against a real Postgres testcontainer.
- [x] Existing filestore unit + integration suites still pass (92 tests, including `test_storage_refactor`, `test_int_filestore`).
- [x] `just prek` passes (ruff, ruff-format, mypy, badge).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,717%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,721%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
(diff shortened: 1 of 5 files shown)


JasperHG90:
Round 3 review came back with one LOW style preference (the `not key` prefix in `is_root_key` is redundant when `key.strip().strip('/')` would also catch `''`). The reviewer themselves notes this is "a style preference, not a bug" and that the explicit short-circuit provides "None-safety defense and readability." Leaving as-is for defensive consistency.

No High/Medium findings remain — closing the review cycle on this branch.

PR #62, opened 2026-04-26
feat(core): atomic note-append endpoint (#56)

### LINKED ISSUE #56: Feature Request: Native Atomic Note Append Endpoint
**Problem:**
The current 'append' workflow for updating notes requires the agent to perform a manual Read-Modify-Write cycle: fetch the full note content, append new text in local context, and push the entire modified body back via `memex_retain` using a `note_key`.

**Context:**
As agents generate long-horizon session logs or iterative reflections, the volume of data being sent back and forth increases linearly. This creates unnecessary token overhead and increases the risk of accidental data loss if the read-modify-write cycle is interrupted or if the agent mismanages the string concatenation.

**Proposed Solution:**
Implement a native `/notes/append` (or similar) endpoint that accepts a `note_id` or `note_key` and a `content` snippet, allowing the server to handle the concatenation atomically. This would significantly reduce the payload size for incremental updates and improve the reliability of session captures.

JasperHG90:
## Summary

Resolves #56 — adds `memex_append_note` across REST / MCP / CLI / Hermes, plus an in-PR fix for a latent overwrite stale-unit bug that the same code paths exposed.

- Server-side atomic delta append: caller sends only the new chunk; server resolves `note_key + vault_id` (or `note_id`), takes `pg_advisory_xact_lock` + `SELECT FOR UPDATE` on the parent, concatenates `parent_body + sep + delta`, and delegates to the existing ingestion pipeline. Two-gate idempotency + incremental block-diff extraction fires naturally — only delta blocks invoke the LLM.
- Idempotent retry: caller-supplied `append_id` keys a new `note_appends` audit table. Replay equality covers `(note_id, delta_sha256, joiner)`; mismatched reuse fails with `409 AppendIdConflictError`. Body update + audit row commit atomically in a single DB transaction.
- In-PR bug fix: incremental overwrite path (`memex_retain` / `memex_add_note` with same `note_key`) now stales `MemoryUnit` rows whose backing chunks were staled by block-diff, mirroring `set_note_status('superseded')` semantics. Append path is regression-guarded never to invoke this cleanup.
- Hermes briefing rewritten to recommend `memex_append` for in-progress session notes; `memex_retain` / `memex_add_note` docstrings demoted with pointers. Telemetry counter surfaces lingering callers to measure migration progress before any future deprecation.

### Surfaces
- `POST /api/v1/notes/append` — `require_write` + explicit vault-access pre-check via `api.notes.resolve_note_id` so vault-leak via `note_key` is impossible.
- `memex_append_note` MCP tool, `timeout=120s`, `tags={'write'}`.
- `memex note append` CLI: `--key/--vault`, `--delta/--delta-file/stdin`, auto `append_id` when omitted.
- `memex_append` Hermes dispatch + `APPEND_SCHEMA` registration.
- Kill-switch `server.append.enabled` (default `true`), 503 + `Retry-After` on lock-acquire timeout.

### Concurrency / safety
- `pg_advisory_xact_lock` keyed on lower 63 bits of parent UUID (collision-bound below any plausible vault size).
- `with_for_update()` added to `set_note_status` so the status/append race window is closed.
- 30s lock-acquire timeout; 503 with `Retry-After` past that.
- NFC-normalized delta hash for retry stability across normalization-divergent intermediaries.
- Frontmatter-prefix regex matches the parser; rejects ambiguous deltas at the schema layer.
- Byte-aware 200 KB delta cap; rejects NUL bytes and whitespace-only deltas.
- `audit_event` emission moved post-commit so rollback can't leak misleading events.

### Migration
`022_note_appends` adds the audit table. New table — no `CONCURRENTLY` needed; downgrade drops cleanly.

## Test plan
- [x] `test_int_note_append.py` — 25 integration tests: happy path (in-place), key/id resolution incl. vault mismatch, idempotent replay, all 3 conflict variants (parent / delta / joiner), 5-way concurrent appends serialise, archived/superseded parents reject, validation (empty / NUL / whitespace / oversized / frontmatter delta), frontmatter-preservation, all 3 joiners, kill-switch (503), lock-acquire timeout (503 with Retry-After), audit rollback on persistence failure, filestore forward recovery
- [x] `test_int_overwrite_stale_units.py` — 5 tests for the in-PR bug fix: stales orphaned units on overwrite, leaves unchanged units active, regression-guards append from triggering staling, no-op when note has zero units
- [x] `test_int_append_route.py` — 16 REST contract tests incl. vault-leak protection (caller with only their own vault gets 403 when targeting another vault's note via `note_key`)
- [x] `test_append_helpers.py` — 21 unit tests for delta validation, hashing, concat
- [x] `test_append_note_mcp.py` — 6 MCP tests via `mcp_client.call_tool`
- [x] 8 new Hermes append tests, 9 new CLI append tests
- [x] `prek run -a` — all hooks pass
- [x] Adversarial-review pass: 3 CRITICAL + 6 HIGH + multiple MEDIUM/LOW findings addressed before opening PR
- [ ] CI green (will verify after open)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,7 +131,7 @@ Embedding and reranking models are configurable via `server.embedding_model` and
 ### Key architectural patterns
 
 - **Distributed reflection queue**: PostgreSQL `SELECT ... FOR UPDATE SKIP LOCKED` for atomic task claiming
-- **Append-only design**: notes are immutable, new versions create new entries (lifecycle: active → superseded/appended/archived)
+- **Note identity is stable; content is mutable in place**. Each note has a stable `note_id` (derived deterministically from `note_key + vault_id`). Two ways to update an existing note's body, both keeping the same `note_id`: (a) **append** — `memex_append_note(note_key=...)` adds a delta to the end (issue #56); (b) **overwrite** — `memex_retain(note_key=...)` / `memex_add_note(note_key=...)` re-ingests the full body. Both run incremental block-diff extraction (only changed chunks invoke the LLM). Lifecycle states (`active` / `superseded` / `archived`) gate operations: appends/overwrites are rejected on non-`active` parents; status transitions are non-destructive. The audit log (`note_appends`, `audit_events`) is genuinely append-only — every mutation leaves a row.
 - **fsspec abstraction**: storage is backend-agnostic (local, S3, GCS)
 - **Circuit breaker**: LLM call resilience with Prometheus metrics
 - **Leader election**: Postgres advisory locks for background reflection scheduling
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,959%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,854%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,854%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,870%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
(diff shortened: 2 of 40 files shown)


JasperHG90:
## Round-4 review disposition (commits 50f8f3e + d7c77a1)

**Addressed**

- HIGH (NFD/NFC test fragility) — replaced literal `'café'` with explicit `unicodedata.normalize('NFD', ...)`/`normalize('NFC', ...)` so the test is robust to any source-file or VCS normalization.
- Med #1 (ingest existence check fetched full body) — switched to `func.length(Note.original_text) > 0` scalar.
- Med #2 (alembic fixture duplication) — extracted shared helpers to `packages/core/tests/integration/_alembic_test_helpers.py`; both alembic test modules import from there.
- Med #3 (redundant `assert isinstance` guards) — removed; `int()` is the type-narrowing.
- Med #4 (silent empty-string fallback for `refreshed_hash`) — added `logger.error` when track_document fails to upsert the hash.

**Disputed**

- The Critical "`NameError` on `uid`" finding does not match the source. `packages/core/src/memex_core/server/notes.py:434` reads `for u in result.get('new_unit_ids') or []` and uses `u` consistently in the body. Verified by inspection on `333d676` and by running the comprehension: it produces the expected `list[UUID]`. Treating this as a Hermes mis-read.

**Deferred**

- Low: CLI `append_note` reads stdin synchronously inside an async command. Consistent with the existing `add_note` pattern — out of scope to change here, would touch all CLI commands.
- Low: Hermes `handle_append` ignores `vault_id` when callers pass it in `args`. Documented behaviour (Hermes is single-vault by design) — adding an explicit warning log is reasonable but a separate concern from this PR.

JasperHG90:
## Round-5 review disposition (commit d7c77a1, plus follow-up 60dad5c)

Round-5 Hermes review (08:55Z): **0 Critical, 0 High, 3 Medium (all wording/doc-only), 5 Low**. No substantive findings; cycle closed.

**Applied** (commit 60dad5c)
- M1 (MCP description "200 KB" misleading) — corrected to "200,000 UTF-8 bytes" so MCP clients don't interpret it as 204,800.

**Deferred** — all are doc-only or refactor suggestions, not bugs:
- M2: comment on `append_extraction_lock_timeout_seconds` already says it covers the extraction phase. Hermes' suggestion to spell out "all internal lock acquisitions, not just the initial advisory" is reasonable but ground-already-covered by the existing comment.
- M3: docstring note that the row-lock vault re-check in `_append_to_note_inner` is the authoritative TOCTOU close. The current code makes this clear via the explicit `if parent.vault_id != parent_vault_id` check.
- L1: `validate_append_delta` runs at both Pydantic and service layers — defense-in-depth for direct API calls bypassing Pydantic. Working as intended.
- L2: `select(Note)` re-query for `user_notes` could be replaced with `session.refresh(parent)`. Minor efficiency, both correct under the row lock.
- L3: implicit contract on `memory.retain` return shape — fragile but pre-existing across the codebase, separate refactor.
- L4: `AsyncMock(side_effect=...)` in one test — minor test cleanup.
- L5: `SET LOCAL lock_timeout` uses f-string with int (safe-by-construction) instead of bindparam — convention nit.

PR ready for merge.

PR #69, opened 2026-04-29
feat(retrieval): D-MEM Z-score embedding anisotropy correction (F2)

JasperHG90:
## Summary
- **F2**: New `AnisotropyCorrector` module implementing D-MEM §4.1 Z-score → sigmoid normalization for cosine similarity scores
- Sliding-window (default k=1024) tracks historical mean and std of pairwise similarities; normalizes new scores through Z-score → sigmoid to counteract embedding anisotropy
- Wired into `RetrievalEngine._compute_pairwise_cosine` so MMR diversity filtering uses corrected similarities
- Cold-start passthrough: returns raw scores unchanged until 32 observations collected
- Config knobs: `anisotropy_window_size` (default 1024), `anisotropy_min_samples` (default 32)
- `AnisotropyCorrectorGroup` for per-site corrector management

## Spec reference
- `cognitive-memory-research-report.md` F2 (D-MEM Z-score anisotropy correction)

## Design note
The anisotropy corrector is applied at the pairwise cosine matrix level (used by MMR), not at the SQL-level `<=>` distance filter. SQL-level filters use hard thresholds; adjusting those requires a different approach (threshold calibration) and is out of scope for F2.

## Test plan
- [x] 19 unit tests covering cold-start passthrough, Z-score normalization behavior, sigmoid bounds, monotonicity, sliding window eviction, running stats accuracy, group API
- [x] All 2244 unit tests pass
- [x] mypy, ruff, ruff-format all pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-3,796%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,799%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-4,091%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-3,796%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/packages/common/src/memex_common/config.py
+++ b/packages/common/src/memex_common/config.py
@@ -710,6 +710,16 @@ class RetrievalConfig(BaseModel):
         default=0.3,
         description='Minimum link weight for causal graph expansion in memory_links.',
     )
+    anisotropy_window_size: int = Field(
+        default=1024,
+        description='Sliding window size for Z-score anisotropy correction. '
+        '1024 matches D-MEM §4.1. Set to 0 to disable.',
+    )
+    anisotropy_min_samples: int = Field(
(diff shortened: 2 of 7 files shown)


JasperHG90:
## Review Findings — Resolution Summary

| # | Sev | Finding | Resolution |
|---|-----|---------|-------------|
| 1 | HIGH | O(n) recompute on every normalize() after window fills | **Fixed**: Replaced `_recompute_stats` fallback with proper eviction tracking — detect `len(window) == window_size` before append, call `_update_stats_remove(evicted)` for the leftmost element. O(1) Welford invariant maintained. `_recompute_stats` removed. |
| 2 | HIGH | `_compute_pairwise_cosine` changed from @staticmethod to instance method | **Safe — no action needed**: Verified no external callers exist outside the class (only `self._compute_pairwise_cosine` at engine.py:495). The change was necessary to access `self._anisotropy`. No subclass overrides found in the codebase. |
| 3 | MED | `AnisotropyCorrectorGroup.get()` not thread-safe | **Fixed**: Added `threading.Lock` to `__init__` and wrapped `get()` in `with self._lock`. Prevents check-then-act race when two threads create duplicate correctors. |

JasperHG90:
## Additional Review Fixes — Pushed in 6d5376f

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | HIGH | O(n) recompute on every `normalize()` after window fills | **Already fixed** in prior commit 8b23f4c: replaced `_recompute_stats` with Welford add/remove, detecting eviction before append |
| 2 | HIGH | `_compute_pairwise_cosine` changed from `@staticmethod` to instance method | **Not a breaking change**: the method is private (`_` prefix) and only called internally via `self._compute_pairwise_cosine`. No external callers exist. |
| 3 | MED | `AnisotropyCorrectorGroup.get()` not thread-safe | **Already fixed** in prior commit 8b23f4c: added `threading.Lock` with `with self._lock` |
| 4 | MED | Warmup observations skew stats while not being normalized | **Deferred**: This is intentional D-MEM behavior — warmup observations contribute to the distribution they'll benefit from once `min_samples` is reached. The alternative (buffering without tracking) would delay convergence. |
| 5 | LOW | Test count regression | **Deferred**: badge update is correct for the current test count |

**New fix in this push:**
| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 6 | LOW | `.replace(tzinfo=utc)` silently overwrites existing timezone | **Fixed**: Added `_to_utc_datetime` helper that uses `.astimezone(utc)` for aware datetimes |

JasperHG90:
## Re-review Fixes — Pushed in fe57957

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | HIGH | `window_size=0` crashes (deque maxlen=0, IndexError) | **Fixed**: Added `_disabled` flag; `window_size=0` returns raw score unchanged. Negative values raise `ValueError`. deque uses `maxlen=max(1, window_size)`. |
| 2 | HIGH | `AnisotropyCorrectorGroup.reset()` iterates without lock | **Fixed**: `reset()` now acquires `self._lock` before iterating |
| 3 | MED | `_compute_pairwise_cosine` signature change | **Not a breaking change**: private method, no external callers |
| 4 | MED | Test count regression | **Deferred**: badge reflects correct current count |
| 5 | LOW | `_m2` floating-point drift | **Fixed**: `self._m2 = max(0.0, self._m2)` after Welford removal |
| 6 | LOW | `.replace(tzinfo=)` timezone overwrite | **Fixed**: Added `_to_utc_datetime` helper (previous commit 6d5376f) |

PR #127, opened 2026-05-02
feat(retrieval): F48 — pre-reranker confidence filter

JasperHG90:
## Summary

Extends F40's hydration `WHERE NOT (...)` predicate with a third OR'd branch: `memory_units.confidence < 0.2`. Strongly-contradicted units get heavily multiplied down by F47's post-reranker boost — pruning them before the cross-encoder reclaims the wasted compute.

## Key decisions (settled across review rounds)

- **Always-on branch**, no separate config flag. `confidence` is `NOT NULL DEFAULT 1.0` (schema invariant), so cold-start units never match the strict `<` 0.2 threshold and SQL three-valued logic does not arise. No COALESCE wrap needed (unlike FSFM).
- **No evidence threshold**. The α-stepping inside the contradiction engine is itself the evidence accumulation — reaching `confidence < 0.2` already requires ≥5 contradict-only events or ≥9 weaken-only events. Adding a count gate would double-count.
- **Single bypass inherited from F40**. Setting `apply_pre_filter=False` drops the entire `WHERE NOT (...)` clause — every branch (MW + FSFM + confidence) is bypassed in one go.
- **F44 exploration bypass already covers F48** — F33's separate hydration path omits the predicate, so low-confidence units pruned by F48 still get re-validation cycles. Confidence cannot become monotonic.

## Pre-merge sequencing

- F40 (#123) — merged on base. Provides the predicate infrastructure + `apply_pre_filter` flag.
- F47 (#125) — merged on base. Validated the confidence ranking signal direction (post-reranker boost) before this PR uses it as a filter.

Both unit and integration test files include a module-level assertion that `apply_pre_filter` exists on `RetrievalRequest`, fail-fast if either dependency disappears.

## Tests

- `packages/core/tests/unit/memory/retrieval/test_f48_sql_builder.py` — 7 SQL-string pinning tests:
  - branch present in SQL when `apply_pre_filter=True` (FSFM off and on)
  - branch absent when `apply_pre_filter=False`
  - OR-joined with existing branches
  - strict `<` (not `<=`)
  - no COALESCE wrap
  - cold-start schema invariant: `NOT NULL DEFAULT 1.0`

- `packages/core/tests/integration/memory/retrieval/test_int_f48_confidence_filter.py` — 4 end-to-end tests across five confidence bands (1.0 / 0.5 / 0.2 / 0.15 / 0.0):
  - main-path prune: <0.2 dropped, ≥0.2 kept (boundary survives strict <)
  - `apply_pre_filter=False` re-surfaces every band
  - vault scoping holds under both pre-filter states
  - F44 exploration self-correction: low-confidence + cold-start unit re-surfaces via F33 bypass

## Test plan

- [x] `uv run pytest packages/core/tests/unit/memory/retrieval/test_f48_sql_builder.py` — 7/7 pass
- [x] `uv run pytest packages/core/tests/integration/memory/retrieval/test_int_f48_confidence_filter.py` — 4/4 pass
- [x] `uv run pytest packages/core/tests/unit/memory/retrieval/test_f40_sql_builder.py` — F40 tests still green (12/12)
- [x] `uv run pytest packages/core/tests/{unit,integration}/memory/retrieval/` (excluding pre-existing test_int_exploration F33 failure) — 493 passed
- [x] `just prek` — ruff + ruff-format + mypy all green

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-4,529%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-4,916%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-4,518%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-4,529%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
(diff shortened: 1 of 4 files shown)


JasperHG90:
Re: Round-1 LOW (seeded_bands fixture commit). Acknowledged. Pattern matches existing integration-test fixtures in this project — `_truncate_db` (autouse) handles cleanup between tests. Not changing this PR.

JasperHG90:
Re: Round-1 LOW (boundary at exactly 0.2). The unit-level SQL pinning test asserts the literal `< 0.2` (strict). The integration test seeds 0.2 and asserts it's kept. Float representation isn't the failure mode — the SQL layer treats 0.2 as a literal in the predicate, not as `0.199999…`. Acceptable coverage; not changing this PR.

PR #130, opened 2026-05-02
feat(retrieval): F42 — reranker variant config (int8 quantization)

JasperHG90:
## Summary

F42 adds opt-in int8 dynamic quantization for the built-in cross-encoder reranker. `OnnxBackend.quantization: Literal['fp32', 'int8']` defaults to `'fp32'` so the feature ships INERT; flipping to `'int8'` lazily produces an int8 ONNX file on first load (cached on disk) for ~2x CPU speedup with low recall risk.

This is the last ticket in the Tier-A latency follow-on slice (after F40 filter, F41 cache, F45 observability). It is the only one with quality-regression risk, hence the default-OFF default and offline-recall measurement design.

## Decision: option (a) — int8 quantization at session creation

The spec offered two paths: (a) int8 quantization or (b) smaller reranker model.

**Picked (a)** because:
- Spec explicitly names ``Int8-batched inference (typically 2x speedup, low recall risk)`` as the primary lever.
- Single config flip; no new HF repo IDs, no second tokenizer, no parallel model registry.
- Production POC on the bundled reranker (``ms-marco-minilm-l12-hindsight-reranker``): weights data 70 MB → 36 MB, sub-0.05 score drift on a 3-pair micro-benchmark.
- Smaller-model would have meant a separate repo per variant and touched more code than F42 warrants.

## Default-OFF rationale

`OnnxBackend.quantization='fp32'` is the default. With no operator action, every existing deployment keeps loading the stock fp32 model with the existing `model_version` shape. F42 cannot regress recall until an operator opts in.

## How recall is measured

**Online (this PR):** new `memex_reranker_latency_seconds` histogram with `variant` label (`fp32` | `int8` | `unknown`). Lets a deployment compare CPU-time per cross-encoder call between variants in Prometheus without a code path swap. Proves the *latency* win.

**Offline (NOT this PR):** the existing eval harness in `packages/eval` is the source of truth for scoring drift / recall@K comparison. F42's online surface only proves the latency lever; recall validation is a separate harness run with `quantization='int8'` set in the eval config. Documented in the integration test docstring.

## F41 cache interaction

`FastReranker.model_version` is now `onnx:<repo_id>:<revision>:<variant>` (`...:fp32` or `...:int8`). F41's score cache keys on `(model_version, query_hash, unit_id)`. Flipping the config from fp32 to int8 produces a different cache key on every entry, so F41 invalidates stored scores naturally on the swap — no bespoke purge required. The integration test asserts `fp32.model_version != int8.model_version` to pin this invariant.

## Reverse path

If recall regresses on a deployment after a flip:

1. Set `server.memory.retrieval.reranker.quantization: fp32` in the operator's config (or unset — `fp32` is the default).
2. Restart the server. The fp32 reranker reloads under the original `:fp32` model_version.
3. No code change needed. The int8 cache file remains on disk; subsequent evaluation runs can reuse it.

## Implementation footprint

- `packages/common/src/memex_common/config.py` — `OnnxBackend.quantization` field.
- `packages/core/src/memex_core/memory/models/quantization.py` — new module, lazy disk-cached int8 quantization helper. Patches `save_and_reload_model_with_shape_infer` *only* during the `quantize_dynamic` call (the bundled model's `word_embeddings` tensor trips ONNX shape inference; the patch is restored on context exit).
- `packages/core/src/memex_core/memory/models/reranking.py` — variant routing in `get_reranking_model`; `FastReranker.variant` property.
- `packages/core/src/memex_core/memory/retrieval/engine.py` — `RERANKER_LATENCY_SECONDS.labels(variant=...).time()` wrapper around `_reranker_score_uncached`.
- `packages/core/src/memex_core/metrics.py` — new `RERANKER_LATENCY_SECONDS` histogram.
- `packages/core/pyproject.toml` — added `onnx>=1.17`. Required at module-load time by `onnxruntime.quantization`. Only imported on the int8 path (lazy import inside the helper).

No changes to `FastReranker.score` / `RerankerModel` protocol / engine boost composition / cache implementation.

## Test plan

- [x] Unit: `packages/core/tests/unit/memory/retrieval/test_f42_reranker_variant.py` (14 cases) — config defaults / validation / int4 rejection, `FastReranker.variant` parsing for fp32/int8/legacy/unknown, get_reranking_model routes to quantization helper only on int8 + threads filename + suffix, fp32 vs int8 model_version differ end-to-end, latency histogram emits via engine path.
- [x] Integration: `packages/core/tests/integration/memory/retrieval/test_int_f42_quantization.py` (2 cases, `@pytest.mark.benchmark`) — real model loading; int8 file caches to disk on first call and short-circuits on second; fp32 vs int8 top-3 ranking agreement >=66% on a 5-doc corpus; both variants' latency is recorded.
- [x] No regressions: 561 common+mcp tests pass; 2598 core unit tests pass (the 4 pre-existing failures on base — `test_alembic_030_resolved_by`, `test_lint_vault_access`, `test_memories_vault_access` — are unrelated and reproduce on the unmodified branch).
- [x] `just prek` clean (ruff + ruff-format + mypy).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-4,935%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-4,951%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/packages/common/src/memex_common/config.py
+++ b/packages/common/src/memex_common/config.py
@@ -287,6 +287,20 @@ class OnnxBackend(BaseModel):
     """Use the built-in fine-tuned ONNX model (default)."""
 
     type: Literal['onnx'] = 'onnx'
+    quantization: Literal['fp32', 'int8'] = Field(
+        default='fp32',
+        description=(
+            'F42: Cross-encoder weight precision. ``fp32`` (default) uses the '
+            'stock model; ``int8`` produces a dynamically-quantised variant on '
+            'first load (cached on disk) for ~2× CPU speedup with low recall '
+            'risk on the supported reranker. The variant flows into '
+            '``FastReranker.model_version`` (``onnx:repo:rev:fp32`` vs '
+            '``onnx:repo:rev:int8``), which keys the F41 score cache so a '
+            'flip naturally invalidates stale entries. Currently only the '
+            'reranker honours this field; on the embedding path it is a '
+            'no-op (logged once at load time).'
+        ),
+    )
 
 
 class LitellmEmbeddingBackend(BaseModel):
--- a/packages/core/pyproject.toml
+++ b/packages/core/pyproject.toml
@@ -26,6 +26,11 @@ dependencies = [
     "httpx>=0.28.1,<0.29.0",
     "langchain-text-splitters>=1.1.0",
     "onnxruntime>=1.23.2",
+    # F42: onnx package supports onnxruntime.quantization.quantize_dynamic for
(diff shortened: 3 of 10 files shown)


JasperHG90:
Closing — F42's fp32↔int8 toggle premise is wrong: the bundled reranker is already trained via QAT (quantization-aware training), so the shipped model IS the int8 inference target. There's no meaningful fp32 baseline to A/B against. F42 will be re-flagged as 'rejected' in BACKLOG with the QAT rationale documented.

PR #134, opened 2026-05-03
feat(extraction): F25b — fold intent+risk classifier into extraction signature

JasperHG90:
## Summary

Combine `ClassifyMemoryUnit` into the existing fact-extraction DSPy signature. One LLM call per chunk instead of one extraction call + one classifier call per fact.

## Why

Direct user request (2026-05-03): "wire the separate DSPy model into the extraction DSPy model. No reason to have separate models here." Halves classifier LLM cost and reduces per-ingest latency.

## What changed

- `extraction/core.py` — `ExtractSemanticFacts` and `ExtractFrontmatterMetadata` docstrings absorb the classifier guidance; intent + risk arrive on each fact.
- `extraction/models.py` — `RawFact` gains `intent_class` + `risk_class` fields with default-on-fail pydantic validators (invalid LLM output → `durable` / `none`); `ExtractedFact` carries them through; `ProcessedFact.from_extracted_fact` populates them.
- `extraction/engine.py` — every `ExtractedFact` construction site copies intent + risk from the `RawFact`; `_classify_and_filter` no longer issues a classifier LLM call (values already on the fact); distribution-metric increments live here.
- `extraction/classifier.py` — keeps `filter_safety_blocked` + `_coerce_intent` / `_coerce_risk`; deletes `ClassifyMemoryUnit`, `_classify_one`, `classify_facts`.
- `metrics.py` — drops `CLASSIFIER_CALLS_TOTAL` (no separate call to count); distribution + blocked counters retained.
- `common/config.py` — `intent_risk_classifier_enabled` flag re-purposed as a kill-switch: when False, every fact is forced to schema defaults regardless of LLM output.
- `tests/unit/.../test_classifier.py` — rewritten around the coerce helpers + RawFact validators + `filter_safety_blocked`.
- `tests/integration/.../test_int_f25b_folded_classifier.py` — new: default-on-fail, safety-filter metric, mocked-LLM round-trip into the DB row, plus a real-LLM golden test (`@pytest.mark.llm`) that verifies Gemini emits valid intent + risk classifications across permanent / ephemeral / private fixtures.
- Old `test_int_classifier_llm.py` (drove `ClassifyMemoryUnit` directly) deleted; the new file's real-LLM test replaces it.
- `cognitive-memory-research-report.md` — §4 F25b entry (six-dimension template).
- `BACKLOG.md` — F25b ticket spec under Tier-A near F25; wave 2 row updated.

## Test plan

- [x] Unit: `pytest packages/core/tests/unit/memory/extraction/` — 426 passed.
- [x] Integration (testcontainers): `pytest packages/core/tests/integration/memory/extraction/ -m integration` — 23 passed (15 deselected as `llm`).
- [x] Real-LLM golden (Gemini, GOOGLE_API_KEY set): `pytest packages/core/tests/integration/memory/extraction/test_int_f25b_folded_classifier.py -m llm` — 1 passed; the model classifies the 4 hand-curated fixtures with intent ≥ 3/4 correct and PII flagged as private/sensitive.
- [x] `just prek` clean (ruff, ruff-format, mypy, prek hooks).
- [ ] Dashboards / alerts referencing `memex_classifier_calls_total` need to be updated (none expected — internal metric).


JasperHG90:
Round 1 / 2 Hermes findings — patch summary + deferred-LOW rationale.

**Patched** (HIGH/MED/MED + trivial LOW):
- `fix(hermes-plugin)` — drop the `classifier.run_dspy_operation` patch site that the F25b refactor invalidated; the `hermes-integration` job was failing on it before any test body ran.
- Round-1 MED — gate `CLASSIFIER_INTENT_DISTRIBUTION` / `CLASSIFIER_RISK_DISTRIBUTION` emission on `intent_risk_classifier_enabled`. Added `TestClassifyAndFilterMetricsGating` covering both paths.
- Round-2 MED-A — retire the dead `_coerce_intent` / `_coerce_risk` helpers; coverage of non-string-garbage shapes preserved on `RawFact`'s validator. Single source of truth now lives on `RawFact`.
- Round-2 MED-B — add `mode='before'` validators on `ExtractedFact.intent_class` / `risk_class` mirroring `RawFact`. Restores the "extraction must never be blocked by a classification mishap" invariant for direct-construction call sites. New `TestExtractedFactClassificationValidators` (incl. an end-to-end `from_extracted_fact` smoke test).
- Round-1/2 LOW (trivial) — drop `async` on `_classify_and_filter` (pure in-process now), and use `RiskClass(...)` in the integration test helper for clarity.

**Deferred LOWs with rationale** (rather than patches):
- *Round-1 LOW: badge count drop 5,011 → 4,600.* This was a snapshot of the F25b commit pre-prek-hook. The auto-update hook ran on the round-0 commit and now reads `5,019` — net +8 after retiring 4 helper tests and adding 13 new tests across the two review rounds. Healthy.
- *Round-2 LOW: `ExtractFrontmatterMetadata` output-model verification.* The signature already declares `extracted_facts: ExtractedOutput`, which wraps `RawFact` — and `RawFact` carries `intent_class` + `risk_class` as native fields with coercing validators. So per-fact intent/risk on the frontmatter path are preserved through the output parser. No change needed.
- *Round-2 LOW: `skip_opinion_formation=True` in the integration test.* That kwarg does not exist on `extract_and_persist` (or anywhere in the extraction surface). Reflection-loop suppression in tests is handled via the `MEMEX_SERVER__MEMORY__REFLECTION__BACKGROUND_REFLECTION_ENABLED=false` env override (see `_set_memex_env_vars` in conftest), not a per-call kwarg. False positive.

JasperHG90:
Round 4 — patch summary + deferred-LOW rationale.

**Patched** (MED):
- Round-4 MED — emit distribution metrics *after* `filter_safety_blocked`, not before. Pre-fix, a `risk_class='safety'` fact got counted on both `CLASSIFIER_RISK_DISTRIBUTION{risk='safety'}` AND `CLASSIFIER_BLOCKED_TOTAL` (double-count) and contradicted the docstring's claim of "what was actually persisted". Now the kept facts are the only ones that land on the distribution; safety-drop telemetry stays exclusively on the dedicated blocked counter. New `test_safety_class_facts_do_not_inflate_distribution` pins the ordering.

**Deferred LOWs with rationale**:
- *Round-4 LOW: validator factory / DRY violation between RawFact and ExtractedFact.* I considered a shared mixin / `validator_factory` but pydantic V2 SQLModel base classes don't compose cleanly across `BaseFact` (SQLModel) and the inline-defined validators — extracting the duplication safely would require an SQLModel mixin pattern that touches the broader fact model hierarchy. That's out of scope for the F25b fold (it touches several non-F25b classes). The two existing 4-line validators are easier to grep for and update together than they are to abstract; the comment-and-mirror pattern is intentional. Defer.
- *Round-4 LOW: per-call set rebuild in `IntentClass`/`RiskClass` literal validation.* `{c.value for c in IntentClass}` is rebuilt in `_classify_and_filter` and in both validator pairs. Hoisting to module-level frozensets is a micro-optimization but introduces an import-time/init-time coupling I don't want to land in a passive PR. Defer until profiling shows it matters (it almost certainly doesn't — these are 3- and 4-element sets rebuilt in already-LLM-bound code paths).
- *Round-4 LOW: `_build_engine` real `dspy.LM` without API key.* Round-3 added a clarifying comment; switching to a `MagicMock`/`None`-typed stand-in is more invasive (the mocked-LLM tests rely on a `dspy.Predict` that's a plain `MagicMock`, but the engine constructor type-annotates `lm: dspy.LM`, so swapping in a `MagicMock(spec=dspy.LM)` is doable but a separate cleanup). Defer.

JasperHG90:
Round 5 — patch summary + deferred-LOW/MED rationale.

**Patched** (MED + 2 LOW):
- Round-5 MED — retire stale ``INTENT_VALUES`` / ``RISK_VALUES`` / ``DEFAULT_INTENT`` / ``DEFAULT_RISK`` exports from ``classifier.py``. Tests now derive these from ``IntentClass`` / ``RiskClass`` directly. ``classifier.py`` is now exactly ``filter_safety_blocked`` plus its module docstring — single-purpose.
- Round-5 LOW (override validation) — reorder kill-switch ahead of override validation. Bad overrides no longer raise when the caller has disabled the classifier at the config level. New ``test_kill_switch_does_not_validate_ignored_overrides`` pins the fast path.
- Round-5 LOW (redundant safety filter) — drop the ``filter_safety_blocked`` call on the kill-switch path. The force-rewrite loop guarantees no fact carries ``risk='safety'``, so the filter was a provable no-op. Skipping it is a trivial perf win on large batches under the kill-switch.

**Deferred (not patching)**:

- *Round-5 MED: kill-switch doesn't prune classification prose from the extraction prompt.* This is real — when the kill-switch is engaged, the LLM still receives the full F25b classification guidance in the ``ExtractSemanticFacts`` / ``ExtractFrontmatterMetadata`` docstrings and still emits ``intent_class`` / ``risk_class`` fields, only for them to be overwritten. The Hermes finding itself notes "Not a correctness bug, but undermines the cost-saving rationale of F25b at the margin." Pruning the prose would mean dynamically rewriting the DSPy signature class docstring at construction time, which DSPy's ``Signature`` API doesn't really support cleanly (signatures are class-level and the prompt template is composed from them statically). A clean fix would either: (a) introduce a separate signature class without the classification fields and dispatch on the kill-switch flag in the engine constructor, or (b) parameterize the docstring via DSPy's instructions API. Both are nontrivial design choices that touch how the engine builds its predictor — out of scope for the F25b fold cleanup. Will track as a backlog enhancement (kill-switch token-budget optimization). The current default (kill-switch off) emits zero waste.
- *Round-5 LOW: ``_build_engine`` un-keyed ``dspy.LM``.* Recurring across rounds 3-5. Round-3 added a clarifying comment; switching to a ``MagicMock(spec=dspy.LM)`` stand-in would change the engine constructor's contract assumptions and the existing mocked-LLM tests rely on the LM being a real object. Defer.

JasperHG90:
Round 6 — patch summary + deferred-LOW rationale.

**Patched** (2 MED):
- Round-6 MED-A — extract ``coerce_intent_class`` / ``coerce_risk_class`` to ``memex_common.schemas`` and route both ``RawFact`` and ``ExtractedFact`` validators through them. Single source of truth for the coercion logic; ``IntentClass`` / ``RiskClass`` value additions can no longer silently desync the two pydantic validators. New ``TestSharedClassificationCoercers`` (6 cases) in ``packages/common/tests/test_schemas.py`` pins the contract.
- Round-6 MED-B — soft-tolerance scoring on the real-LLM risk fixture in ``test_real_llm_extraction_emits_valid_intent_and_risk_per_fact``. Replaced the per-fixture hard assert with ``risk_correct >= risk_total // 2`` (floor 1), matching the intent fixture's best-of-N pattern. Eliminates flakiness on single-shot LLM runs without losing the "did the model flag PII somewhere" signal.

**Deferred LOWs**:
- *Round-6 LOW: ``logger.warning`` when overrides are silently discarded by kill-switch.* Pure observability suggestion. The kill-switch fast-path I added in round-5 is the right place for it, but turning silent discard into a warning is a UX call that interacts with how often the kill-switch is used in production (every ingest call would log if the kill-switch is on AND any caller passes overrides — could be noisy). Defer to a follow-up that adds the warning gated behind a "verbose" config knob.
- *Round-6 LOW: ``_build_engine`` un-keyed ``dspy.LM``.* Recurring across rounds 3-6. Same rationale as before — switching to a ``MagicMock(spec=dspy.LM)`` would change engine constructor's contract assumptions and the real-LLM test branch in the same file currently relies on a real ``dspy.LM`` being constructible. The round-3 comment is the smallest defensible response. Defer.

This was the 6th iteration. Per orchestrator's standing rule (cap at 6 rounds, defer LOWs), no further patches are planned. Both round-6 MEDs are addressed.

JasperHG90:
Round 8 — at the round-6 cap. Two MEDs remain; both are deferred per orchestrator rules. No further patches.

**Deferred (out of scope / architectural-input-required)**:

- *Round-8 MED-1: kill-switch metric absence is itself a signal.* Hermes recommends a dedicated ``memex_classifier_kill_switch_active`` gauge so operators can distinguish "no facts processed" from "kill-switch engaged." This is a real observability point but it's adding *new* metric instrumentation, not fixing a bug — and the fix Hermes itself acknowledges as reasonable (suppress distribution metrics under kill-switch) was the round-1 patch. Adding a new operator-facing gauge is an observability-design call beyond the F25b fold scope. Track as a backlog enhancement: kill-switch operator visibility.

- *Round-8 MED-2: ``coerce_intent_class`` / ``coerce_risk_class`` rebuild enum value sets on every call.* Recurring finding (LOW in rounds 4/5/6/7, escalated to MED in round 8). Module-level frozenset caching would shave a few µs off each fact's validation. Hermes round-7 itself called this "negligible" perf-wise; the validators run inside an LLM call path that already costs seconds per fact. Module-level mutable state at import time also has subtle test-isolation implications (other tests already mock out enum members in some suites). The micro-optimization is real but the fix lives outside the F25b fold's stated scope of "remove the second LLM call per fact, halve classifier cost." Track as a backlog enhancement: frozenset-cache the validator value sets.

**Round-8 LOWs deferred without comment** (per orchestrator standing rule):
- Override-validation enum rebuild (companion to MED-2).
- ``ProcessedFact.from_extracted_fact`` re-parse from str→enum (data-model design note for future rework, not a bug).
- Real-LLM test fixture coverage suggestion (add a second non-trivial-risk fixture).

This is iteration 8 of the review cycle (1 conftest fix + 6 patch rounds + 1 trivial PEP-8 round). Per orchestrator's cap-at-6 rule, escalating to the orchestrator for the remaining MEDs.

PR #136, opened 2026-05-03
feat(agent-surface): F3 — codify 4-layer memory-routing primer across MCP/Hermes/Claude Code/CLAUDE.md

JasperHG90:
## Summary

Codify the 4-layer memory routing (Episodic / Semantic / Conceptual / Procedural-observations) across all four agent-facing surfaces. Pure docs/prose change — no behaviour change, no new schema, no new tools.

Source: `cognitive-memory-research-report.md` §2.3 (memory-types mapping) + §4 F3 (6-dimension contract). Same shape as F43.

## What changed

- `packages/mcp/src/memex_mcp/_f3_descriptions.py` — new canonical primer constants (prose / markdown table / fragment).
- `packages/mcp/src/memex_mcp/server.py` — appended layer primer to the 5 search-tool descriptions: `memex_memory_search`, `memex_note_search`, `memex_get_entity_mentions`, `memex_kv_search`, `memex_survey`.
- `packages/hermes-plugin/.../memex/briefing.py` + `templates.py` — `_LAYER_ROUTING_PRIMER` next to F43's `_RESOLUTION_FLOW_PRIMER`; `LAYER_ROUTING_PROMPT_FRAGMENT` added to `__all__`.
- `packages/claude-code-plugin/rules/memory-layers.md` — new rule file (auto-installed by `on_session_start.sh`'s existing `rules/*.md` glob — no hook change).
- `packages/claude-code-plugin/skills/{recall,remember}/SKILL.md` — reference the new rule.
- `AGENTS.md` (the `CLAUDE.md` symlink target) — new "Memory layers and tool routing" section pointing at the three canonical sources.
- `packages/hermes-plugin/tests/test_f3_layer_primer_parity.py` — cross-surface parity test (10 cases).
- `cognitive-memory-research-report.md` §4 F3 — marked implemented (2026-05-03).
- `BACKLOG.md` — F3 ticked.

## Design choices

- **Canonical text lives in `memex_mcp._f3_descriptions`** for the MCP surface. Hermes-plugin does not depend on `memex-mcp` (`packages/hermes-plugin/pyproject.toml`) so the briefing / templates / rule / `CLAUDE.md` surfaces all duplicate the canonical text — the parity test pins them together. This mirrors F43.
- **Three constants** (prose / table / fragment) so each surface uses the format that fits its rendering context (compact prose for tool descriptions; markdown table for human-readable briefings/rules; concise fragment for Hermes prompt templates).
- **No `__init__.py` re-export from `memex_mcp`** — kept private (leading underscore) like the other `_f*_descriptions` modules.

## Out of scope

No new schema, no new tools, no new layer. Core / Cross-Context layers (per §2.3 ZenBrain expansion) stay informational. F3 codifies the existing implicit mapping — see report §2.3.

## Test plan

- [x] `uv run pytest packages/hermes-plugin/tests/test_f3_layer_primer_parity.py -x` — 10/10 pass.
- [x] `uv run pytest packages/hermes-plugin/tests/` — 419/419 pass (no regression).
- [x] `uv run pytest packages/mcp/tests/` — 414/414 pass (no regression, including F43 parity).
- [x] `just prek` — ruff + ruff-format + mypy + test-count-badge all green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -200,7 +200,7 @@ Claude Code rule file (parity required per the agent-surface rule).
 | **Episodic** ("what happened, when") | Timestamped, source-attributed Notes — sessions, reflections, decisions | `memex_note_search` / `memex_recent_notes` / `memex_find_note` | "Find yesterday's reflection about the deploy regression" |
 | **Semantic** ("decontextualised facts") | MemoryUnits — short fact/observation/event statements extracted from notes | `memex_memory_search` / `memex_get_memory_units` / `memex_get_entity_mentions` | "What does v2 use for auth?" |
 | **Conceptual** ("synthesised mental models") | MentalModels — reflection output bundling per-entity observations with trend tracking (new/strengthening/stable/weakening/stale) | `memex_survey` / `memex_get_entities` (with `mental_models=True`) | "What do you know about Project X overall?" |
-| **Procedural-observations** ("adaptations to context") | KV entries under `procedure:<verb>:<context-tag>` — observations about how to adapt your existing skills, NOT the procedures themselves | `memex_kv_search` / `memex_kv_get` with `prefix='procedure:'` | "For this user, `deploy` means staging — never prod after 6pm" |
+| **Procedural-observations** ("adaptations to context") | KV entries under `procedure:<verb>:<context-tag>` — observations about how to adapt your existing skills to a context, NOT the procedures themselves | `memex_kv_search` / `memex_kv_get` with `prefix='procedure:'` | "For this user, `deploy` means staging — never prod after 6pm" |
 
 Default to `memex_memory_search` for content-shaped questions and
 `memex_note_search` for source-shaped questions. The agent owns the verb;
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -188,6 +188,27 @@ Async mode: `asyncio_mode = "auto"` — all async tests run automatically.
 - Ensure `ensure_db_env_vars` fixture is active for E2E tests.
 </constraint>
 
+## Memory layers and tool routing
+
+Memex stores four memory layers; pick the right tool for the layer you
+need (canonical source: research report §2.3 + §4 F3). The mapping is the
+same across MCP tool descriptions, the Hermes session briefing, and the
+Claude Code rule file (parity required per the agent-surface rule).
+
+| Layer | What it stores | Retrieve with | Tiny example |
+|---|---|---|---|
+| **Episodic** ("what happened, when") | Timestamped, source-attributed Notes — sessions, reflections, decisions | `memex_note_search` / `memex_recent_notes` / `memex_find_note` | "Find yesterday's reflection about the deploy regression" |
+| **Semantic** ("decontextualised facts") | MemoryUnits — short fact/observation/event statements extracted from notes | `memex_memory_search` / `memex_get_memory_units` / `memex_get_entity_mentions` | "What does v2 use for auth?" |
+| **Conceptual** ("synthesised mental models") | MentalModels — reflection output bundling per-entity observations with trend tracking (new/strengthening/stable/weakening/stale) | `memex_survey` / `memex_get_entities` (with `mental_models=True`) | "What do you know about Project X overall?" |
+| **Procedural-observations** ("adaptations to context") | KV entries under `procedure:<verb>:<context-tag>` — observations about how to adapt your existing skills, NOT the procedures themselves | `memex_kv_search` / `memex_kv_get` with `prefix='procedure:'` | "For this user, `deploy` means staging — never prod after 6pm" |
+
+Default to `memex_memory_search` for content-shaped questions and
+`memex_note_search` for source-shaped questions. The agent owns the verb;
+Memex owns the adverb. Authoritative copies live in
+`packages/mcp/src/memex_mcp/_f3_descriptions.py`,
+`packages/hermes-plugin/.../briefing.py` `_LAYER_ROUTING_PRIMER`, and
+`packages/claude-code-plugin/rules/memory-layers.md`.
+
 ## When the user reports an issue resolved (§3.5 5-step flow)
 
 When the user says "the X bug is fixed" / "we shipped Y" / "issue Z is no
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -191,9 +191,11 @@ Async mode: `asyncio_mode = "auto"` — all async tests run automatically.
 ## Memory layers and tool routing
 
 Memex stores four memory layers; pick the right tool for the layer you
-need (canonical source: research report §2.3 + §4 F3). The mapping is the
-same across MCP tool descriptions, the Hermes session briefing, and the
-Claude Code rule file (parity required per the agent-surface rule).
+need (canonical spec: research report §2.3 + §4 F3). The mapping is the
+same across all five agent surfaces — MCP tool descriptions, the Hermes
+session briefing, the Hermes templates fragment, the Claude Code rule
+file, and this section — enforced by `test_f3_layer_primer_parity.py`
+(parity required per the agent-surface rule).
 
 | Layer | What it stores | Retrieve with | Tiny example |
 |---|---|---|---|
@@ -204,10 +206,12 @@ Claude Code rule file (parity required per the agent-surface rule).
 
 Default to `memex_memory_search` for content-shaped questions and
 `memex_note_search` for source-shaped questions. The agent owns the verb;
-Memex owns the adverb. Authoritative copies live in
-`packages/mcp/src/memex_mcp/_f3_descriptions.py`,
-`packages/hermes-plugin/.../briefing.py` `_LAYER_ROUTING_PRIMER`, and
-`packages/claude-code-plugin/rules/memory-layers.md`.
+Memex owns the adverb. The single source of truth for the primer strings
+lives in `packages/common/src/memex_common/agent_surface.py`; consumers
+(`packages/mcp/.../server.py` via the `_f3_descriptions.py` shim,
+`packages/hermes-plugin/.../briefing.py` and `.../templates.py`,
+`packages/claude-code-plugin/rules/memory-layers.md`, and this section)
+all import or mirror that one module — never re-declare the text.
 
 ## When the user reports an issue resolved (§3.5 5-step flow)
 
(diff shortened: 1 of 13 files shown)


JasperHG90:
## Round 1 — Hermes findings response

**HIGH #1 (CLAUDE.md not modified):** False positive. `CLAUDE.md` is a symlink to `AGENTS.md` (`readlink CLAUDE.md` → `AGENTS.md`), and `AGENTS.md` does carry the §"Memory layers and tool routing" table — see lines 191–209 of the diff. The parity test reads `_REPO_ROOT / 'CLAUDE.md'` and resolves through the symlink, so the assertions pass on CI (verified locally: 12/12 in `test_f3_layer_primer_parity.py`). No patch needed for this finding, but I did surface a real drift while addressing HIGH #2 — see below.

**HIGH #2 (two table copies can drift):** Patched. `LAYER_ROUTING_PRIMER_*` strings now live in `memex_common.agent_surface` (a new shared module). MCP, Hermes briefing, and Hermes templates all import from there. `memex_mcp._f3_descriptions` is now a thin re-export shim, and the new `test_table_primer_single_source_of_truth` / `test_prompt_fragment_single_source_of_truth` tests pin object-identity (`is`, not `==`) so a future split would fail before any column-level drift. While wiring up the full-row assertion I caught a real pre-existing drift: `AGENTS.md` was missing `to a context` in the Procedural-observations row. Fixed.

**MED #1 (`PROMPT_FRAGMENT` vs `PRIMER_FRAGMENT` naming):** Patched via the same SoT refactor — `templates.LAYER_ROUTING_PROMPT_FRAGMENT` is now a re-export of `agent_surface.LAYER_ROUTING_PRIMER_FRAGMENT`. The templates.py-side name is preserved for back-compat (matches the `*_PROMPT_FRAGMENT` suffix convention used by sibling fragments in that file), but it's the same string object.

**MED #2 (`asyncio.run()` per parametrized case):** Patched. `_surfaces()` is now a `@pytest.fixture(scope='module')`, so the event-loop spin-up happens once instead of 8× across the parametrized layer × tool tests.

**MED #3 (misleading `verbatim` test name):** Patched. Renamed to `test_table_surfaces_carry_full_canonical_rows` and assert the **complete** canonical row content (layer name + "What it stores" + "Retrieve with" + "Tiny example" columns), not just row prefixes. The four canonical rows are pinned in `_CANONICAL_TABLE_ROWS`, and the test first asserts each row appears verbatim in `LAYER_ROUTING_PRIMER_TABLE` before checking each surface — so drift between the constant tuple and the source of truth fails too.

**LOW #1 (badge regression 5,049 → 4,630):** Resolved organically. The pre-commit `update test count badge` hook ran during this round and bumped the badge to 5,061 (reflecting the new tests added in this branch + recent main). The 4,630 number was a stale auto-update that pre-dated the F3 wave's tests.

JasperHG90:
## Round 2 — Hermes findings response

**HIGH (CLAUDE.md surface):** Patched. You're right that this kept reading as "missing surface" in static review — the issue was that `CLAUDE.md` is a `readlink CLAUDE.md` → `AGENTS.md` symlink (single inode on disk), which a diff-based reviewer can't see. I've changed the parity test to read `AGENTS.md` directly (`_AGENTS_MD_PATH`), and added a docstring + inline comment calling out the symlink relationship. This is functionally equivalent (same bytes are loaded) but lets future reviewers — and Hermes — see the surface that actually carries the table.

**MED (stale `templates.py` comment):** Patched. The F3 block now correctly points at `memex_common.agent_surface.LAYER_ROUTING_PRIMER_FRAGMENT` instead of the now-obsolete `memex_mcp._f3_descriptions` reference.

**MED (`is`-only check fragile under module reload):** Patched. Both SoT identity tests now assert `==` (content equality — graceful under any future module-reload edge case) AND `is` (object identity — the primary structural check that pins the SoT pattern). The test also now imports the canonical primers from `memex_common.agent_surface` directly, which is a cleaner architectural statement than going through the `_f3_descriptions` re-export shim.

**LOW (over-long lines + `# noqa: E501`):** Skipping — stale advice. The project ignores `E501` globally (`pyproject.toml [tool.ruff] lint.ignore = ["E501"]`), and ruff passes prek as-is. Adding `# noqa: E501` would be redundant on these lines.

**LOW (inconsistent import path — server.py via shim, others direct):** Skipping. The asymmetry is intentional: `memex_mcp.server` lives in the same package as `_f3_descriptions.py`, so going through the shim is the natural same-package import; cross-package consumers (`memex_hermes_plugin`) import directly from `memex_common.agent_surface` (the actual source of truth). I've updated the test to follow the cross-package convention so the architectural statement is consistent for any reader looking at the test as the "canonical example".

JasperHG90:
## Round 4 — Hermes findings response

**MED #1 (`test_mcp_search_tools_carry_prose_primer` re-runs `asyncio.run`):** Patched. New `mcp_tool_descriptions` module-scoped fixture computes the per-tool dict once via a single `asyncio.run(...)` call; both the prose-primer test and the `surfaces` fixture consume it. No second event loop, no risk of nested-loop conflict.

**MED #2 (`memory-layers.md` "byte-for-byte" claim is misleading):** Patched. Reworded the canonical-sources footer to say "row-level verbatim parity" and qualify with the specific test path (`test_table_surfaces_carry_full_canonical_rows`). Matches what the test actually enforces (four canonical layer rows, not whole-file equality).

**LOW #1 (`_f3_descriptions.py` docstring — "historical names" / back-compat false rationale):** Patched. Rewrote the docstring: this module is a **same-package re-export shim** that lets `memex_mcp.server` import primer strings from a sibling MCP-package module instead of scattering `from memex_common.agent_surface import ...` lines through the route handlers; cross-package consumers (`memex_hermes_plugin`) skip the shim entirely.

**LOW #2 (server.py diff context can't show `@mcp.tool(name=...)` for `memex_survey` / `memex_kv_search`):** Skipping — speculative diff-readability nudge that the parity test makes moot. `test_mcp_search_tools_carry_prose_primer` runs against the `_F3_TOOLS` tuple (which explicitly names `memex_survey` and `memex_kv_search`), fetches each tool from the registered MCP server via `mcp.get_tool(name)`, and asserts the prose primer is present in every description. If a tool is mis-registered or the primer wasn't appended, the test goes red. Diff-context expansion would be belt-and-suspenders against a failure mode the test already catches.

JasperHG90:
## Round 5 — Hermes findings response

**MED (no SoT identity test for MCP shim):** Patched. `test_mcp_shim_single_source_of_truth` now asserts `is`-identity (with `==` check skipped here because object identity is the *only* failure mode this test cares about; content-equality drift is already covered by the substring + full-row tests). It iterates all three primer constants (`PROSE`, `TABLE`, `FRAGMENT`) on `memex_mcp._f3_descriptions` and verifies each one is the same Python object as the corresponding `memex_common.agent_surface` constant. The diagnostic names the specific constant if drift occurs, so a future regression is debuggable from the failure message alone.

**LOW (`from __future__ import annotations` inconsistency in `agent_surface.py`):** Skipping. Round 3 explicitly flagged the `from __future__ import annotations` in `agent_surface.py` as redundant (no type annotations in the module) and asked it be removed; round 5 now flags its absence as inconsistent with sibling files. Removing it is the correct call for a string-only module — happy to revisit if you'd prefer the consistency framing instead, but I don't want to flip-flop the file across rounds.

**LOW (`_LAYER_ROUTING_PRIMER` alias in `briefing.py` is unnecessary):** Skipping. The alias preserves the existing module-level naming convention for primer constants in `briefing.py` (`_RESOLUTION_FLOW_PRIMER`, `_STORAGE_MODEL_PRIMER`, `_ROUTING_GUIDE`). Consistent with sibling primers; the `is` test makes the indirection cheap to verify.

JasperHG90:
## Round 6 — Hermes findings response (final round)

**MED (`asyncio.run` could conflict with pytest-asyncio auto-mode):** Skipping — premise is incorrect. The hermes-plugin test suite is **`asyncio: mode=Mode.STRICT`**, not `auto` (verified locally: `pytest -v` header reads `asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None`). The `asyncio_mode = "auto"` cited from AGENTS.md is the root-level memex test configuration; the hermes-plugin overrides it (intentional, since the plugin's contract with Hermes Agent is sync). Under STRICT mode, no event loop is running outside `@pytest.mark.asyncio`-marked tests, so the `RuntimeError: This event loop is already running` failure mode this finding describes cannot occur for a sync module-scoped fixture. The pattern `asyncio.run(...)` inside a sync fixture has been stable across 5 prior CI runs on this branch alone.

**LOW (`agent_surface.py` raw-string note / no escape sequences):** Acknowledged. No action needed — the markdown table is plain content with no backslash sequences, and the comment is informative for future maintainers. (Hermes itself notes "currently clean".)

**LOW (`_f3_descriptions.py` docstring says "four" surfaces, actually five):** Acknowledged. Off-by-one cosmetic — the actual scope is "the four Python-importable surfaces share strings via this shim; the fifth (`AGENTS.md`) is a markdown surface kept in parity by the test suite." Not patching to keep round 6 minimal.

**LOW (`_CANONICAL_TABLE_ROWS` line lengths):** Skipping — same as round 2: the project ignores `E501` globally (`pyproject.toml [tool.ruff.lint] ignore = ["E501"]`), and `ruff` passes prek as-is. A `# noqa` would be redundant.

**LOW (`importorskip('memex_mcp')` over-gates non-MCP tests):** Skipping — speculative refactor. In practice `memex_mcp` is always installed as part of the workspace (it's a sibling package consumed by the integration tests), so the importorskip never actually skips in this repo. Splitting into two test files for a hypothetical "memex_common-only consumer" trades a one-line guard for two-file coupling and a divided test surface, which I judge a worse tradeoff for the actual constraints.

**LOW (PEP 8 import group for `briefing.py`):** Acknowledged. ruff-format ran clean and is the project's authoritative formatter; the visual separation between cross-package (`memex_common`) and same-package (`.async_bridge`) imports is intentional. Not patching.

**LOW (`templates.py` 43-char import alias indirection):** Acknowledged. Hermes itself confirms the SoT identity test passes — the alias serves the SoT pattern (round 1's design choice) and the F3 block comment explains why; not changing in round 6.

---

This is round 6, the orchestrator-imposed cap. The remaining MED is speculative (rooted in a wrong test-config premise) and the LOWs are non-blocking. Proceeding with merge.

PR #137, opened 2026-05-03
feat(cli): F7 — memex lint review interactive CLI

JasperHG90:
## Summary

`memex lint review` — human-side parity verb for F8's agent-driven lint triage. Walks each pending finding interactively, renders evidence + suggested action, lets the user accept / dismiss / skip via single keypress. Defaults to dry-run; requires `--apply` to write.

## What changed

- `packages/cli/src/memex_cli/lint.py` — new `@app.command('review')` subcommand.
- `packages/cli/src/memex_cli/lint_review.py` — new module: per-finding render + prompt loop.
- Tests: unit (11 cases) + integration (3 cases) with audit-log-invariant check.
- §4 F7 marked as shipped in research report.
- BACKLOG.md: F7 ticked.

## Out of scope

- No MCP / Hermes / Claude Code parity (F7 is human-only by design — F8 is the agent counterpart).
- No new lint check types.
- No destructive default.

## Test plan

- [ ] Unit + integration tests pass (audit-log invariant verified).
- [ ] Manual smoke: `memex lint review --vault X` → renders findings, accepts single keypress, `--apply` writes; without `--apply` no DB changes.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -98,7 +98,7 @@ Total scope: ~22-30 weeks for Tier S+A (incl. ~2-day Wave 0) + ~3-4w for Tier-A 
 - [x] **F6** — Maintenance ledger + rule-based linter (4-6w) — runs under existing `MEMEX_LEADER_LOCK_ID` — see report §4 F6 (shipped 2026-05-01 via memory_augmentation)
 - [x] **F8** — `memex_get_lint_flags` MCP (1-2w) — depends on F6 — see report §4 F8 (shipped 2026-05-01 via memory_augmentation)
 - [x] **F10** — Surprise-gated LLM lint (3-5w) — depends on F2 + F6; LLM cost-capped per vault — see report §4 F10 (shipped 2026-05-01 via memory_augmentation; polarity discrimination deferred to F10b)
-- [ ] **F7** — `memex lint review` interactive CLI (1-2w, promoted 2026-05-02) — see report §4 F7 + Issue #18 Comment 1. Human-side parity verb for F8's agent-driven lint triage; analogous to `git add -p`. **Concrete scope:**
+- [x] **F7** — `memex lint review` interactive CLI (1-2w, promoted 2026-05-02; shipped 2026-05-03) — see report §4 F7 + Issue #18 Comment 1. Human-side parity verb for F8's agent-driven lint triage; analogous to `git add -p`. **Concrete scope:**
   - Extend `packages/cli/src/memex_cli/lint.py:27` with a new `@app.command('review')` subcommand alongside the existing four (`status`, `findings`, `dismiss`, `resolve`).
   - Optional new module `packages/cli/src/memex_cli/lint_review.py` for the per-finding render + prompt loop, kept narrow so the four shipped subcommands stay simple.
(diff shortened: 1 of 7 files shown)


JasperHG90:
Re: Hermes round-1 LOW — *Global scope filters client-side after fetching all vaults' findings*

Deferring to a follow-up. The behaviour is intentionally consistent with the existing `lint_findings_cmd` — adding a `vault_id__isnull` filter param is a server/API surface change that crosses route + client + service layers, out of scope for F7's CLI-only deliverable. Tracking on the BACKLOG follow-up list rather than patching here.

JasperHG90:
Re: Hermes round-2 LOWs — both deferred to follow-ups, neither patched here.

**LOW: ``--global --limit 50`` may return < 50 due to client-side filter applied after the limit-capped fetch** — confirmed, but the fix is a server-side ``vault_id__isnull`` filter param spanning route + client + service layers, same scope as the round-1 LOW comment above. F8 (the agent counterpart) and a separate refactor PR are the right venue.

**LOW: README badge dropped from 5,062 → 4,646** — investigated. The drop is **not** an F7 regression. ``pytest --collect-only`` (which the pre-commit ``update test count badge`` hook reads) currently fails with 3 collection errors on this worktree:
- ``packages/hermes-plugin/tests`` — ``ModuleNotFoundError: No module named 'memex_hermes_plugin'``
- ``packages/mcp/tests/test_int_get_memory_units_chunk_ids.py`` — duplicate basename collision with ``packages/core/tests/integration/test_int_get_memory_units_chunk_ids.py``
- ``packages/mcp/tests/test_intent_risk_validation.py`` — same duplicate-basename pattern

Those uncollectable tests are silently dropped from the count, so the badge under-reports relative to ``release/tier-a-cognitive-memory``. F7's own contribution to the count is +2 cases (the new round-1 + round-2 tests). Worth a tracker for ``conftest.py`` / package-install hygiene but out of scope for the F7 CLI deliverable.

JasperHG90:
Re: Hermes round-3 findings.

**MED (re-raised) — README test count drop**: standing by the round-2 reply ([linked above](https://github.com/JasperHG90/memex/pull/137#issuecomment-4365789703)). The drop is caused by 3 pytest collection errors on this worktree (``hermes-plugin`` import failure + 2 duplicate-basename collisions in ``packages/mcp/tests``) — pre-existing, unrelated to F7. The badge value F7 produces (4,646 → 4,647 with the new id-guard test, after this round's commits) is the **correct** count of *collectable* tests in this environment. Fixing the collection errors is out of scope for a CLI feature PR; tracking as a separate cleanup task. Not patching here.

**MED — global + limit truncation**: Patched as a docstring + ``--limit`` help-text update (``e70dfd68``). The actual server-side ``vault_id IS NULL`` filter remains a multi-layer follow-up.

**LOW — missing ``id`` defensive guard**: Patched (``0b72f9c0``).

**LOW — ``_prompt_verdict`` default-label cosmetic**: deferred. The ``(s)`` suffix Typer appends is genuinely redundant with the ``[s]kip`` label, but it's a pure UX nit and changing it would either (a) hide the default (worse UX) or (b) require a custom prompt renderer (over-engineering). Leaving as-is.

PR #141, opened 2026-05-03
feat(retrieval): F22 — Two-Factor edge confidence (variance alongside mean)

JasperHG90:
## Summary

Extends F47's confidence-boost composition to certainty-aware form: `1.0 + α × (confidence − 0.5) × certainty` where `certainty = 1 − variance/MAX_VARIANCE`. New `confidence_evidence_count` column tracks negative-evidence accumulation; Beta(1, 1) posterior gives cold-start neutrality. Lint gating, reflection prioritisation, and F33-style edge_exploration all benefit from the new signal.

## What changed

- Migration 033: `confidence_evidence_count` column with backfill (contradicts+weakens links only — reinforces excluded for symmetry with forward path).
- New `memory/confidence.py` — pure-Python Beta(1,1) posterior helpers (`mean_and_variance`, `certainty`, `sample_concentration`).
- Contradiction engine: bump count on weaken/contradict, untouched on reinforce.
- F47 composition extended via `certainty_modulation_enabled` (ship default False — F47 boost is bit-for-bit unchanged at ship time).
- Lint gating: skip findings on high-variance / cold-start units (rule-based + LLM lint).
- Reflection cron: prioritise high-variance edges.
- F33 edge_exploration mode: epsilon-greedy injection of uncertain edges.
- `MemoryUnitDTO.confidence_variance` computed-derived field.
- `CONFIDENCE_VARIANCE_OBSERVED` histogram.
- §4 F22 marked shipped; F22 ticked.

## Ship-time defaults (no behaviour change)

- `certainty_modulation_enabled = False` → F47 boost preserved bit-for-bit.
- Migration ships column + backfill inertly.

## Three independent kill-switches

1. `certainty_modulation_enabled = False` → F47 baseline.
2. Revert contradiction-engine bump → count stays at backfill values; F47 still works.
3. Migration revert → drop column.

## Test plan

- [x] All 5 BACKLOG numerical cases pass (cold-start, single observation, well-evidenced contradicted, midpoint, clean).
- [x] `MAX_VARIANCE == 1/12` pinned.
- [x] Backfill: contradicts+weakens counted; reinforces excluded.
- [x] Forward path: count bumps on weaken/contradict only.
- [x] Regression guard: `certainty_modulation_enabled=False` → F47 boost bit-for-bit identical regardless of evidence_count.
- [x] Schema CHECK rejects negative `confidence_evidence_count` at the database.
- [x] `just prek` passes (ruff, ruff-format, mypy).
- [x] 56 F22 tests pass (30 unit confidence + 9 unit certainty modulation + 9 unit lint gating + 4 integration forward path + 4 integration retrieval composition).

## Known v1 limitation (documented)

Reinforcement-only units characterised as high-variance (count=0). Documented mitigation: monitor `CONFIDENCE_VARIANCE_OBSERVED` segmented by reinforces-count category; ship symmetric backfill+forward extension if disproportionate exploration budget accumulates on these units.


JasperHG90:
## PR-review-cycle round 7 — deferral notes

Round 6 cap reached. Round-7 review surfaced 2 MEDs and 2 LOWs. Triage:

**MED-A (per-unit UPDATE loop)** — addressed in c353fcc7 with a `TODO(perf, post-v1)` comment. Hermes itself flagged this as "Fine for v1 given typical contradiction-batch sizes, but worth a TODO(perf) if scaling is expected".

**MED-B (`_process_flagged_unit` plain tuple return → suggest NamedTuple)** — **deferred**. The tuple is an internal contract (single caller, `_detect`). Two recent signature revisions (round-4 HIGH, round-5 HIGH) landed cleanly via positional unpacking. NamedTuple would be a nice-to-have but it's a defensive refactor against a hypothetical future signature change, not a fix for an observed bug. Re-raisable in a follow-up PR if a third revision in this area surfaces.

**LOW-1 (`_coerce_metadata_to_dict` str-as-Mapping)** — deferred. The JSONB column never returns a `str`; the hazard is theoretical. The `isinstance(..., (str, bytes))` guard is a one-liner if it ever surfaces.

**LOW-2 (test backfill SQL idempotency)** — deferred. The integration test exercises the canonical zero-start path; idempotency-with-non-zero-start is a separate test scenario worth adding to the F22 follow-ups list.

**LOW-3 (`LintConfidenceGate.variance_max` `le=1/12` literal)** — deferred. A future Beta-prior change is already a deliberate-edit affair (see `MAX_VARIANCE` pin test); the two-place coupling is documented by the F22 module docstring's "single-prior convention" framing.

The round-7 verdict line was: **"Overall: solid implementation."** Merging.

JasperHG90:
## Round-8 deferral — reinforce asymmetry MED

The reinforce-asymmetry observability gap (round-8 MED on `contradiction/engine.py`) is the **documented v1 limitation** in BACKLOG: forward/backfill symmetry is the design choice — `reinforces` events do NOT bump `confidence_evidence_count`, mirroring the migration's exclusion. The `CONFIDENCE_VARIANCE_OBSERVED.observe(variance)` call in the rerank path now fires unconditionally (round-8 fix above), giving operators the variance distribution. Segmenting the histogram by `reinforces`-link-count to surface well-reinforced cold-evidence units is the explicit follow-up the BACKLOG already enumerates as out-of-scope for v1 — extending the symmetry to reinforces is a deliberate v2 decision, not a regression in this PR.

LOWs from round 8 (warning verbosity, deferred imports, parameter-convention asymmetry, `confidence_variance` default, missing concurrent-batch test) all defer to follow-ups.

JasperHG90:
Re: round-11 LOW on `test_f22_certainty_modulation.py` Prometheus internals.

Treating this as a no-op for this round. The `_sum.get()` delta pattern is safe under our actual test execution model:

- `tool.pytest.ini_options` has no xdist config; tests run sequentially in a single process.
- `asyncio_mode = "auto"` runs async tests on a per-test event loop, not concurrently.
- Hermes itself flags this as a hypothetical: "running under pytest-xdist (separate processes) is safe; running in-process with concurrent test classes would not be."

If we ever add xdist, the per-test-class delta pattern would still be process-isolated. The only concrete risk is in-process concurrent classes, which pytest doesn't do without explicit threading. No code change.

JasperHG90:
Re: round-12 LOW on \`from typing import Any\` in \`confidence.py\`.

Dismissing as not actionable. \`Any\` IS used at line 47 (\`def extract_confidence_and_count(unit: Any) -> tuple[float, int]\`). With \`from __future__ import annotations\`, the annotation is a forward-string at runtime, but the symbol still has to be importable for static type checkers (mypy) and IDE introspection. The codebase imports \`from typing import Any\` in 47 files for the same reason — this is the convention. Moving \`Any\` under \`if TYPE_CHECKING:\` would be a stylistic choice, not a correctness fix, and inconsistent with the rest of the package.

JasperHG90:
**Round-14 MED (alembic 033 backfill SQL — redundant JOIN): verified-incorrect, kept as-is**

Hermes flagged the inner ``JOIN memory_units mu_inner ... AND mu_inner.confidence_evidence_count = 0`` as redundant with the outer ``mu.confidence_evidence_count = 0``. After verifying the loop semantics, this finding is incorrect:

- The inner SELECT picks the next ``_BACKFILL_BATCH_SIZE`` unit_ids ordered by ``to_unit_id``.
- Without the inner filter, every iteration re-picks the SAME first ``batch_size`` unit_ids in the contradicts/weakens link set.
- The outer filter rejects all updates on the second iteration, ``rowcount=0``, the loop terminates after only the first batch lands.
- The inner filter is what advances the candidate window through the link target set as each batch flips its targets' ``confidence_evidence_count`` non-zero.

I committed an explicit in-code note (``7366aaca``) recording this conclusion so the same finding doesn't recur in a future review.

JasperHG90:
**Behavioural change — reinforce-event accumulation (Hermes round-16 MED, surfaced for changelog)**

This PR fixes an event-loss bug in the contradiction engine's confidence-update path. The PR body update is currently blocked by a GitHub Projects (classic) GraphQL deprecation error, so I'm posting the changelog note as a comment for visibility.

**Prior behaviour:** ``confidence_updates`` was an absolute-value dict keyed by unit_id, so multiple events on the same target unit in one ``_detect`` batch were last-writer-wins (one event survived).

**Post-fix:** ``confidence_deltas`` accumulates signed alpha-step deltas, so all events land.

- For weaken/contradict (monotonic-decreasing) the post-batch endpoint is identical — the SQL ``GREATEST/LEAST`` clamp absorbs the difference.
- For reinforce (monotonic-increasing) the endpoint is strictly higher post-fix: two reinforces on the same target previously landed at ``confidence + α``; they now land at ``confidence + 2α`` (clamped to 1.0 by the ``LEAST(1.0, …)`` guard).

Operators with reinforce-heavy workloads will see ``memory_units.confidence`` move strictly higher post-deploy. This is the intended bug-fix behaviour. See in-code note at ``contradiction/engine.py:147-178``.

JasperHG90:
**Round-17 MED (alembic 033 backfill GROUP BY undercount): verified-incorrect, kept as-is**

Hermes worried that ``GROUP BY ml.to_unit_id ... LIMIT :batch_size`` could undercount if a single ``to_unit_id``'s group spans multiple index pages. This is incorrect — Postgres SQL semantics guarantee that ``GROUP BY`` produces one row per group with the full count, regardless of the planner's chosen strategy:

- HashAggregate: reads all matching rows into a hash table, emits one row per group with the full count.
- GroupAggregate: requires input sorted by the GROUP BY column; once a new group key arrives, the prior group's full count is emitted.

In either strategy the LIMIT is applied to the *aggregate output* (one row per unit_id), not to the *input scan* (one row per link). A single unit_id never gets split across batches — its ``cnt`` is always its full count, by SQL semantics.

We also have an existing integration test (``test_int_f22_confidence_composition.py::test_backfill_counts_contradicts_and_weakens_only``) that asserts on exact counts (5, 0, 0) for three units across mixed link types — if the GROUP BY were splitting groups, this test would fail.

No code change needed.

JasperHG90:
**Round-24 dismissals (verified against actual source)**

**HIGH 2 — `logger.error` argument mismatch in migration 033** is a false finding.

Verified at `033_confidence_evidence_count.py:312-325` (post round-24 push, head SHA `97eab1b3`):

```python
logger.error(
    'F22 backfill verification: %d units have ...'
    'with the contradiction engine paused.',
    mismatch_count,   # ← exactly ONE positional arg, ONE %d placeholder
)
```

The format string contains exactly ONE `%d` placeholder, and the call passes exactly ONE positional formatting arg (`mismatch_count`). The trailing comma on the argument list (legal Python) was misread by the reviewer as a second omitted positional. There is no second arg; `mismatch_count` IS interpolated correctly.

Round-24 patches landed for the genuine findings:
- HIGH 3 (NaN/inf clamp guards) — fix(confidence,lint-confidence)
- MED 4 (DTO upper-bound documentation) — docs(schemas)
- MED 5 (`_MAX_VARIANCE` constant in config.py) — refactor(config)

MED 6 (str→UUID type hint) is cosmetic; skipped per "no new API surfaces" constraint.

JasperHG90:
## Round-28 response

All HIGH and CRITICAL findings from round 26 are resolved (module-scoped RNG, driver-independent rowcount, `int()` cast on mismatch verification). This round has no HIGH/CRITICAL findings — two MEDIUMs and five LOWs.

### MEDIUM findings — deferred with rationale

**1. MAX_VARIANCE triplication**: The three sites are cross-referenced with inline comments pointing to all locations, and `TestDtoFormulaConsistency` pins equivalence. A `memex_common.constants` module is the right follow-up but is out of scope for this PR (it requires updating imports across packages/core and packages/common, and both packages have circular-import constraints that need resolution first). The triplication is safe because the test catches any drift at CI time.

**2. Migration race not DB-guarded**: The migration docstring explicitly documents the operational requirement (pause the contradiction engine). An advisory lock inside the migration would couple the migration to the application's lock-ID scheme and break for zero-downtime deployments where the app is already paused. The `MEMEX_MIGRATION_FAIL_ON_MISMATCH` env-var provides CI/CD actionability (set to `1` in CI), and the `logger.error` with aggregate count provides deploy-log observability. A `SELECT pg_advisory_lock` guard is a reasonable follow-up for deployments that can't pause the app.

### LOW findings

**3. `_rng` fork() seed sharing**: Valid under `fork()` without `exec()` (gunicorn preload). The retrieval engine uses `self._rng` (instance-scoped) for its own ε-greedy roll, but `exploration.py` is module-scoped. The fix (accept a seed parameter or use `os.register_at_fork`) is a follow-up — the impact is duplicate exploration sequences in forked workers, not a correctness bug.

**4. `_clamp_confidence_pair` bool handling**: Already fixed in the current code — `isinstance(evidence_count, int) and not isinstance(evidence_count, bool)` at line 65. Hermes reviewed a stale diff.

**5. Metadata mutation**: Documented thoroughly in both injector docstrings with explicit "cross-function hazard" callouts. The retrieval engine's documented call-site invariant (disjoint pools, single-use per request) makes this safe.

**6. `object.__setattr__` on confidence**: Intentional defense-in-depth. The `Field(ge=0, le=...)` constraint fires during construction for user-supplied values; the `object.__setattr__` path is for computed-assignment within the validator where the inline bounds check is the guard.

**7. Config weight message**: Unrelated to this PR (pre-existing in `ReflectionConfig`).

---

All findings are documented trade-offs, already-fixed issues, or follow-up material. No code changes needed.

PR #151, opened 2026-05-05
refactor(cli): standardize vault flags + nest eval CLI benchmarks

JasperHG90:
## Summary

- **memex CLI**: Every vault-accepting command now uses `--vault`/`-v` with `str | None = None`, defaulting to the user's configured active vault. No command requires an explicit vault anymore. Help text standardized across all commands.
- **memex-eval CLI**: Restructured from flat `locomo-*` commands + top-level `run` to `memex-eval <BENCHMARK> <SUBCOMMAND>` pattern. `run` → `internal run`, `locomo-*` → `locomo <subcommand>`. Deprecated backward-compat aliases added for old `locomo-*` commands.

### memex CLI vault flags (8 files)
- `memory.py`: deprioritize, restore, reconsolidate, consolidate — changed from required `str` to `str | None = None` with default to `config.vault.active`. All resolve via `api.resolve_vault_identifier()`. Search help text standardized.
- `diagnose.py`: manifold, retrieval, summary, lint — changed from required `str` to `str | None = None`, added `-v` short flag, default resolution.
- `consolidate.py`: tick, status — added `-v` short flag, standardized help text.
- `lint.py`: status, findings, review — added `-v` short flag, standardized help text.
- `notes.py`: add, search, list, recent, find, export — standardized help text. `append` — dropped conditional `--key`-requires-`--vault` check, now uses `config.write_vault` fallback.
- `entities.py`: delete-mental-model — renamed param from `vault_id` to `vault`.
- `session.py`: briefing — standardized help text.

### memex-eval CLI restructuring (1 file)
- Created `internal_app` sub-Typer, moved `run` from `@app.command()` to `@internal_app.command('run')`
- Created `locomo_app` sub-Typer with 6 nested commands: ingest, export, answer, judge, report, efficiency
- Added 6 deprecated backward-compat aliases on root `app` with `deprecated=True` + `warnings.warn()`
- LongMemEval unchanged (already nested)
- Updated test: `['run', '--help']` → `['internal', 'run', '--help']`, LoCoMo test paths updated

## Test plan
- [x] All existing CLI tests pass (81 key tests verified)
- [x] `memex memory deprioritize --help` shows `-v, --vault` with "Defaults to the active vault"
- [x] `memex diagnostics manifold --help` shows `-v, --vault`
- [x] `memex-eval internal run --help` works
- [x] `memex-eval locomo --help` shows nested subcommands
- [x] `memex-eval locomo-ingest --help` shows deprecation notice
- [x] `memex-eval run` returns "No such command"
- [x] ruff + ruff-format + mypy all pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-5,115%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-5,116%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
 <<<<<<< HEAD
-  <img src="https://img.shields.io/badge/tests-5,549%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-5,115%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,6 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-<<<<<<< HEAD
   <img src="https://img.shields.io/badge/tests-5,115%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
--- a/packages/cli/src/memex_cli/consolidate.py
+++ b/packages/cli/src/memex_cli/consolidate.py
@@ -36,7 +36,7 @@ async def tick_cmd(
     ctx: typer.Context,
     vault: Annotated[
         str | None,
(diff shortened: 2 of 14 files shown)


JasperHG90:
Addressed in c7d308a:
- **HIGH** `entities.py:delete_mental_model` — now calls `resolve_vault_identifier(vault)` instead of `parse_uuid`, matching the help text.
- **HIGH** `notes.py:append_note` — replaced the `config.vault.active` mutation with the same local-variable + `config.write_vault` fallback pattern used everywhere else. No side effects on the shared config object.

Notes on the remaining MEDIUM findings:
- **`list[str]` vault args claim "names or UUIDs"** — verified accurate. The HTTP server resolves names via `resolve_vault_ids()` for `/notes` (list/recent), `/notes/find`, and inside `SearchService.search_notes` (via `_vaults.resolve_vault_identifier(str(v))`). End-to-end name support works.
- **`None` propagation to `resolve_vault_identifier`** — when no `--vault` and no active vault is configured, the call raises `VaultNotFoundError`, which surfaces in the CLI as a 404 with a clear message and a `memex vault list` suggestion via `handle_api_error`. The existing error path is user-friendly.
- **Removed `--key` + `--vault` guard in `append_note`** — intentional. Acknowledged.

Skipping the LOW findings (deprecated alias boilerplate, `deprecated=True` Typer version) — explicit signatures keep the deprecation contract auditable, and `deprecated=True` works in the pinned Typer version (verified locally).

CI failures in `python-tests (3.12)` (F14/F43 SKILL.md content tests) are pre-existing on the base branch — unrelated to this PR.

JasperHG90:
Second-pass review (independent agent, since Hermes hit quota) caught three more real issues — fixed in 117c4be:

- **`entities.py:delete_mental_model`** — was passing `vault_id=None` when no `--vault`, letting the *server's* default vault take over instead of the client's `config.vault.active`. Now matches the diagnose/memory pattern. Vault resolution also moved ahead of the confirmation prompt so a typo'd vault name fails before the user confirms a destructive op.
- **`notes.py:add_note`** — same `config.vault.active = vault` mutation Hermes flagged HIGH on `append_note`. The fix-commit only patched `append_note`; `add_note` retained the side effect. Replaced with the same local-variable + `config.write_vault` fallback. Updated 3 downstream `config.write_vault` reads to use the local. (The `effective_vault` assignment is placed after input validation so the file-not-found early-exit path doesn't NPE on a missing config.)
- **`lint.py`** — help text "Defaults to the active vault." lied. `_resolve_scope` actually defaults to `'all'` (every scope) when no flag is given. Corrected to "Filter to one vault by name or UUID."

Tests: `test_add_note_with_vault` was previously a *snapshot of the bug* (asserting `config.vault.active` got mutated) — rewritten to assert the vault flows into the request DTO and config is NOT mutated. Added `test_deprioritize_no_vault_configured_surfaces_error` to lock down the negative path. 700/702 pass (2 deselected = pre-existing F32 failure on the base branch, unrelated).

The independent reviewer's other findings (deprecated alias warning visibility, eval CLI test coverage, and a few Lows) are scope-creep relative to this PR's stated goal — flagging for follow-up rather than including here.

PR #158, opened 2026-05-08
feat(cc-plugin): converge with Hermes — auto-capture, vault resolution, namespaced KV

JasperHG90:
## Summary

Brings the Memex Claude Code plugin in line with the Hermes plugin's design choices where the divergence wasn't intentional. Six new behaviors land together because they share state files and hook plumbing — splitting them would create an awkward intermediate state.

### Capture trigger (converges with Hermes)
- **`SessionEnd` auto-capture** ingests the full transcript via `memex_add_note(background=true)`. Gated on `reason ∈ {prompt_input_exit, logout, other}`; skipped on `clear`/`resume` so a continuing session isn't double-captured.
- **`PreCompact` active append** writes pre-compaction turns to the same `session:<id>` note via `memex_append_note`. The first event creates the note; subsequent events extend it in place. **One CC session = one note**, not a series of fragments.
- This is a safety net **under** `/remember`, not a replacement: explicit captures stay the curator's tool; auto-capture mines context the human didn't flag.

### Vault resolution + KV namespace (converges with Hermes)
- New hierarchical resolver: `project → user → agent → MEMEX_VAULT → server default`. Mirrors `packages/hermes-plugin/.../project.py:8-14`. In-process TTL cache.
- KV writes now use `app:claude-code:*` namespace (mirrors `app:hermes:*`) with **one-time forward-migration** from bare `project:<id>:vault` keys. Legacy keys stay in place per the user-only KV deletion policy.

### Auto-tag injection
- `PreToolUse` hook on `mcp__memex__memex_add_note` enriches every call with ambient metadata: `surface:claude-code`, `session:<id>`, `project:<id>`, `git:branch=*` / `git:sha=*` / `git:repo=owner/name`, `git:dirty` (presence-only), `claude:model=*`, `cc:plugin=*`. Git tags are **omitted** (not "unknown") when context isn't resolvable.
- Defaults `background=true` and `vault_id=<active>` when not explicitly set. Caller-supplied values are preserved; tags merge-not-replace. Critically uses an explicit `if $c == null` check rather than jq's `//` operator so an explicit `background: false` survives.

### Other plumbing
- **Transcript-aware `/recall`** (`UserPromptExpansion` hook) composes a query from the last N turns when `/recall` is invoked without arguments. N defaults to 3, configurable via `MEMEX_CC_RECALL_TURNS`, clamped to `[1,10]`.
- **`track_capture.sh`** (`PostToolUse` hook) bumps a per-session counter (atomic via `O_APPEND`) for future safety-net heuristics.
- **`memex` CLI calls** wrapped in `timeout` (default 8s, `MEMEX_CC_TIMEOUT` override, validated `[1,600]`) so a hung server can't block hooks. Especially important for the `async` SessionEnd hook.

### Robustness work (informed by adversarial review)
- jq pipelines explicitly handle `content: null`.
- Transcript shrinkage detection: if the file shrunk below the recorded offset, reset and re-capture.
- `transcript_path` is checked for readability, not just existence.
- KV write failure during migration logs to stderr.
- `resolve_config.sh` no longer emits `systemMessage` JSON unless the caller opts in via `MEMEX_RESOLVE_VERBOSE=1`, so PreToolUse hooks aren't polluted.

### Process

3 rounds of adversarial review by an independent reviewer, citing file:line for every claim:
- **Round 1**: 17 issues found → 12 fixed, 5 declined with verified rationale (race conditions in declined cases were verified non-issues per CC hook lifecycle).
- **Round 2**: 1 medium issue (`MEMEX_CC_TIMEOUT` validation) → fixed.
- **Round 3**: 1 minor (no test for the timeout fix) → 9 parametrized test cases added. Verdict: "VERDICT: YES, MEETS 99% CERTAINTY BAR. Confidence: 99%."

## Test plan

- [x] **Unit tests (83 cases)** under `packages/claude-code-plugin/tests/unit/` — PATH-shimmed mock CLI, threading-based concurrency test, parametrized timeout validation, edge cases (null content, transcript shrinkage, unreadable transcript, missing/null tags, malformed JSONL, wrapped envelopes).
- [x] **Integration tests (5 cases)** under `packages/claude-code-plugin/tests/integration/` — drive every argv shape the scripts emit through the real Memex Typer CLI (API mocked at the `get_api_context` boundary, mirroring `tests/test_cli_*.py`). Catches flag-name and arity drift between scripts and CLI.
- [x] **E2E tests (3 cases)** in `tests/test_cc_plugin_e2e_lifecycle.py` — real Postgres testcontainer + FastAPI + Alembic. Parametrized KV namespace round-trip, cross-plugin namespace isolation (`app:claude-code:*` vs `app:hermes:*`).
- [x] **Full suite**: 95 tests pass in ~17s.
- [ ] **Real-session smoke test**: install plugin into a clean CC instance, drive `/remember`, force a compaction, exit; confirm a single `session:<id>` note exists with the expected tag set and full transcript.
- [ ] **Backward compatibility**: install over an existing plugin install with bare `project:<id>:vault` KV entries; confirm forward-migration writes the new key on first SessionStart.

## Files

- **Implementation** (10 files): all under `packages/claude-code-plugin/scripts/` plus `hooks/hooks.json`. Helper `memex_persist_session_delta` lives in `resolve_config.sh` so PreCompact + SessionEnd share it.
- **Docs** (4 files): `skills/{remember,recall,retro}/SKILL.md` + `README.md`.
- **Tests** (13 files): `_helpers.py` + `conftest.py` + `fixtures/{mock_uvx,mock_memex}` + 7 unit test files + 1 integration file + 1 E2E file at the project root.
- **Build** (1 file): `pyproject.toml` adds `packages/claude-code-plugin/tests` to `pythonpath` so `_helpers.py` is importable.

## Out of scope (filed as follow-up)

- Port the rich auto-tag scheme to Hermes' `_ingest_session_note` so notes from either plugin carry uniform metadata.
- Optional `/create-specialist` skill for vault-bound subagents.
- Live notes / `source_query` re-runs (backend feature; not a plugin concern).


JasperHG90:
Hermes round-5 follow-up:

- **HIGH ("`on_post_commit.sh` / `on_post_write.sh` are unverified")** — false positive. Both scripts predate this PR:
  - `on_post_commit.sh` was added in `75ce5a0c` (P1–P8 hook integration)
  - `on_post_write.sh` was added in `08bb3b2e` (hook hardening)
  - hooks.json had matchers pointing at both well before this branch was cut
  - This PR's `hooks/hooks.json` change only adds `description` fields next to the existing entries (`6af502df`)

  No code change in this PR introduces, modifies, or moves the post-commit/post-write hooks — they're out of scope and stay covered by their original commits' test suites.

- **MEDIUM 2 + MEDIUM 3** (`grep -E` regex injection in `mock_memex`, missing `returncode` assertions in `test_active_vault_falls_back_to_{user,agent,env_var}`) — fixed in `9534b9d3`.

- The remaining MEDIUMs (4: `mock_uvx` argument parsing; 5: `MEMEX_RESOLVE_VERBOSE` double-emit; 6: redundant outer `*.md` loop) and all LOW items were explicit "minor"/"no action needed" by Hermes itself; not patching to keep the review iteration converging.

JasperHG90:
Round-7 follow-up:

- **HIGH-1** (git ls-remote on every hook): cached in `$STATE_DIR/refcache/<sanitised-ref>` with a 24h TTL via `find -mmin +1440`. `latest` users are unaffected (no validation ever runs).
- **HIGH-3** (temp-file/EXIT-trap race): withdrawn by Hermes — confirmed.
- **HIGH-4** (macOS lacks `timeout(1)`): added a `gtimeout` probe in `memex()`.
- **HIGH-5** (`mock_uvx` argument parsing): test-infrastructure fragility on a flag combination we don't emit. Not patching to keep diff surface minimal.
- **HIGH-2** (`trap 'echo "{}"; exit 0' ERR` swallows internal errors silently): this is intentional and load-bearing for hooks that **must never block the Claude Code host**. Real bugs in tag injection / jq pipelines surface in the **unit tests** (96+ now exercise the exact code paths) rather than in production stderr. Adding diagnostic logging in the trap risks (a) chatter on Apple devices where stderr is shown to the user, and (b) a debug-log file under `$STATE_DIR` that nobody reads. If a future need emerges, `MEMEX_CC_DEBUG=1` + `set -x` redirection is a one-line opt-in we can ship then.
- **MEDIUM-6** (tag array `${_tags[@]}` quoting): withdrawn by Hermes — confirmed.
- **MEDIUM-7 / MEDIUM-9 / MEDIUM-10 / MEDIUM-11 / MEDIUM-12 / MEDIUM-13** and **LOW-14**: all explicitly "fine"/"consistent"/"justified" per Hermes. Not patching.
- **MEDIUM-8** (SSH URLs with port leaked `:22` into the project ID): patched; new parametrize entry.

PR #168, opened 2026-05-12
feat(retrieval): log-additive composition with configurable clip (V19)

JasperHG90:
## Summary

Replaces the five-factor multiplicative composition in retrieval reranking with a log-additive form bounded by a configurable clip `L`:

\`\`\`
final_score = ce_score * exp(clip(sum(log(b_i)), -L, +L))
\`\`\`

- Ships **default `L = math.inf`** — a *mathematical no-op* (within 1e-9 of the prior \`ce * b_recency * b_temporal * b_mw * b_confidence * b_decay\` product for strictly positive boost inputs).
- Adds \`composite_boost_log_clip\` to \`RetrievalConfig\` with Pydantic \`ge=0.0\` validation (rejects negatives + \`-math.inf\`).
- Adds two Prometheus surfaces: \`memex_composite_boost_clipped\` (Histogram — at L=∞ directly observes the pre-clip aggregate product, subsuming the V17 standalone histogram plan) and \`memex_composite_boost_non_finite_guard_triggered_total\` (Counter — short-circuit indicator for NaN/±inf inputs).

The empirical \`L\` lands in a follow-up config commit using the histogram's distribution. **V17 is a prerequisite for that follow-up only, not for this code merge.**

## Links

- **Ticket**: V19 in [SPRINT.md](https://github.com/JasperGinn/memex/blob/release/tier-a-cognitive-memory/SPRINT.md) (gitignored locally; tracked in the working tree)
- **Plan**: \`~/.claude/plans/memex-composition-log-space-clip.md\` (private, off-tree)
- **Design doc**: \`DESIGN_DOCUMENT.md\` retrieval-scoring section (gitignored)

## What changed

| File | Change |
|---|---|
| \`packages/common/src/memex_common/config.py:876\` | New \`composite_boost_log_clip: float = math.inf\` field |
| \`packages/core/src/memex_core/memory/retrieval/engine.py:85-139\` | New \`_compose_boosts_logspace\` helper + \`LOG_FLOOR_COMPOSITE_BOOST\` |
| \`packages/core/src/memex_core/memory/retrieval/engine.py:1690-1700\` | Compose-call replaces the multiplicative product |
| \`packages/core/src/memex_core/metrics.py:174-189\` | \`COMPOSITE_BOOST_CLIPPED\` + \`COMPOSITE_BOOST_NON_FINITE_GUARD_TRIGGERED\` |
| \`packages/core/src/memex_core/memory/retrieval/engine.py:1623\` (descriptions) | Five per-factor alpha descriptions rewritten to drop the "Multiplicative" claim |
| \`packages/core/tests/unit/memory/retrieval/test_composition_log_clip.py\` | 39 unit cases: identity at L=∞, finite-L bounds, zero/negative boost handling, NaN guard, non-finite guard, config validator, rank-preservation |
| \`packages/core/tests/unit/memory/retrieval/test_reranking_boosts.py\` | \`TestCompositeLogClipIntegration\` (8 cases) driving the full \`_rerank_results\` entry point including a true L=0 ↔ L=∞ discriminator test |
| \`packages/core/tests/benchmarks/test_retrieval_benchmarks.py\` | New informational \`composition\` benchmark group |
| \`docs/explanation/retrieval-scoring.md\` | New: log-additive rationale, L-value bounds table, floor justification |
| \`docs/how-to/server-configuration-templates.md:126\` | New \`composite_boost_log_clip\` row + drops "Multiplicative" from \`reranking_recency_alpha\` row |
| \`README.md\` | Test-count badge regenerated to 6,322 (matches collection under full extras+groups sync) |

## Surfaces

Not touched. Same retrieval API; same MCP / CLI / plugin surfaces. Score *values* change at sub-1e-9 floating-point level; rank ordering preserved (E7 baseline is the cross-cutting gate).

## Pre-PR gates

- **Self-assessment: 99%.** Gap: -1% for three LOW findings outside V19's diff (pre-existing stale \`engine.py:1581-1588\` anchors in \`docs/how-to/evaluation-suite.md\` and \`packages/eval/internal_evaluations.md\`, plus pre-existing stale "multiplicative" wording in \`test_f47_confidence_boost.py:3\` and \`test_int_f11_decay_composition.py:43-46\`). Pre-existing hygiene gaps V19 surfaced but did not introduce; folded into the V19 follow-up cleanup list.
- **Adversarial review loop**: 5 rounds (rounds 5–9 in this branch's commit history). Final round (9) returned zero CRIT/HIGH/MED. Each round patched by a separate commit; fresh reviewer each round; no cycle cap honored.

## Test plan

- [ ] \`uv run pytest packages/core/tests/unit/memory/retrieval/test_composition_log_clip.py packages/core/tests/unit/memory/retrieval/test_reranking_boosts.py -v\` → 67 passing
- [ ] \`uv run pytest packages/core/tests/benchmarks/test_retrieval_benchmarks.py::test_composition_logspace -v\` → informational, no regression vs informational baseline
- [ ] E7 baseline ranking gate at \`L = math.inf\` → drift ≤ tolerance (no-op proof)
- [ ] \`/metrics\` scrape on a dev server → \`memex_composite_boost_clipped\` and \`memex_composite_boost_non_finite_guard_triggered_total\` series appear after a query
- [ ] Anchor verifier on the plan file → exit 0, all anchors \`ok\` (confirmed: 3/3 ok)


JasperHG90:
## Response to Hermes round 1 (https://github.com/JasperHG90/memex/pull/168#pullrequestreview-_kwDORU_WIs7-nJIH)

### MED — `math.inf` JSON serialization → **fixed** in commit 77c26682

You were right that this had teeth. `RetrievalConfig` is transitively serialized to JSON via the `/system/config` endpoint (`server/system_routes.py:42` calls `MemexConfig.model_dump(mode='json')`). Verified empirically that Pydantic v2 was silently emitting:
- `mode='json'` dict: float `math.inf` (then `json.dumps` turns it into `Infinity` — RFC-8259-invalid)
- `model_dump_json()`: JSON `null` (silent data loss; reload would fail `ge=0.0`)

Fix: paired `@field_validator(mode='before')` + `@field_serializer(when_used='json')` on `composite_boost_log_clip`. Wire form is now the standard string `"inf"`; Python value stays `math.inf`. Negative L rejection via `ge=0.0` is unchanged. Nine regression tests added in `TestConfigJSONSerialization`, including a strict RFC-8259 compliance assertion that would have failed pre-fix (`parse_constant=lambda c: raise`).

### LOW 1 — negative `log_clip` not guarded inside `_compose_boosts_logspace`

Leave as-is. The helper is module-private (single underscore + module-internal only — `engine.py:1690-1700` is the lone caller, and it reads `composite_boost_log_clip` from a Pydantic-validated `RetrievalConfig` instance with `ge=0.0`). The validator already rejects negative L and `-math.inf` at construction time, so the function can never receive a negative `log_clip` from a real call path. An internal-guard layer would either be unreachable defensive code or an `assert` that production code would strip. The non-finite guard (the only one that *can* fire) is justified because non-finite floats can flow in from upstream model outputs (cross-encoder overflow, NaN propagation from a corrupted score), which the config layer can't catch.

### LOW 2 — `ce_score=inf` short-circuit makes the unit unbeatable

Leave as-is — intentional. `ce_score=inf` is an upstream cross-encoder bug (sigmoid overflow or model error); making the unit "unbeatable" is the conservative behavior because we don't want the metadata clip to silently mask a broken score. The `COMPOSITE_BOOST_NON_FINITE_GUARD_TRIGGERED` counter fires on this path, so the condition is observable. Adding a `logger.warning` here would log on every reranking call when the upstream is broken — that's a log-storm pattern rather than the right observability surface. The histogram + counter pair is the right level. (If a future incident shows the counter alone is insufficient, the upstream cross-encoder normalization layer is the better place to log.)

### LOW 3 — benchmark prose doesn't separate math from metric-emission overhead

Acknowledged. The 20-30× slowdown vs the bare-product reference is dominated by Prometheus histogram `observe()` (atomic bucket update), not the `log`+`exp` math. The benchmark is intentionally end-to-end (math + observability) because that's what production actually runs; a math-only variant would not reflect the deployed cost. Will fold a note into the benchmark docstring on the next PR that touches that file — not patching in this commit to keep the diff scoped.

---

Self-assessment unchanged: **99%**, with the same -1% gap acknowledged in the PR body (pre-existing stale anchors in eval docs and test docstrings outside V19's diff). The MED fix lands the no-op-default contract on solid ground for the entire serialization surface, not just the in-memory path.

JasperHG90:
## Response to Hermes round 2 (https://github.com/JasperHG90/memex/pull/168#pullrequestreview-_kwDORU_WIs7-bug)

### MED — `log_clip = -inf` slipped through the guard → **fixed** in commit 29874625

You caught a real correctness bug: `math.isnan(-math.inf)` is False, so the negative-inf path bypassed the NaN check and `max(-(-inf), min(-inf, log_boost))` collapsed to `+inf`, multiplier `= exp(inf) = inf`. Config's `ge=0.0` blocks this from real call sites, but the function's own contract was promised in the docstring ("negative L is rejected") and not enforced standalone — a future refactor moving the function or a direct test caller could have hit it.

Fix: added `log_clip < 0` to the short-circuit. Function is now correct independent of caller validation. Four new tests in `TestNegativeLogClipGuard` cover: finite-negative input, -inf input, counter increment, histogram skip. Docstring updated to call out the defense-in-depth motivation.

### LOW — `_serialize_composite_boost_log_clip` sign precision → **fixed**

Switched `math.isinf(value)` to `value == math.inf`. Single-sign match; if -inf ever reached the serializer (impossible via `ge=0.0` but defensive), it now passes through as a float instead of round-tripping as +inf.

### LOW — `_parse_composite_boost_log_clip` return type → **fixed**

Widened annotation from `object` to `float | object`. Communicates the numeric-on-success contract at the type level; Pydantic's coercion is unchanged.

### LOW — `parse_constant` deprecated → **fixed**

You're right that the assertion is redundant: the serializer emits JSON strings (`"inf"`) never bare `Infinity` literals, so the test's intent is satisfied without `parse_constant`. Replaced with direct substring checks: no `Infinity`/`NaN` tokens, explicit `"composite_boost_log_clip":"inf"` literal. Removes the 3.14 deprecation footprint.

### Suggestion — explicit NaN config-level test → **added**

`test_nan_rejected` in `TestConfigValidator`. NaN already fails `ge=0.0` (NaN comparisons return False), but the test documents the contract.

### Suggestion — lower histogram buckets (0.02, 0.05) → **deferred**

The current buckets `(0.1, 0.3, ..., 50.0)` cover the realistic aggregate range `[~0.03, ~7.6]` for `[0.5, 1.5]` factors. Adding 0.02 / 0.05 is genuinely useful when tuning toward a finite `L` (so the operator can see what the clip is squeezing out of the tail), but at ship default `L = math.inf` there's nothing to tune yet. Holding for the empirical-L follow-up config commit where bucket resolution at the chosen tail becomes load-bearing. Filed as a follow-up note.

---

Self-assessment unchanged: **99%**. Test count now 84 in the V19 suite (was 79 pre-round-2; +5 for negative-clip guard + NaN config rejection).

JasperHG90:
## Response to Hermes round 4 (loop exit gate cleared)

Round 4 returned **0 CRIT / HIGH / MEDIUM**, 4 LOWs. The pre-PR adversarial loop's exit criterion is met. CI green on cef24043: pre-commit, python-tests, mcp-import-check, hermes-integration, all 4 build-python jobs, docker-build-memex.

### LOW disposition

**L1 — undifferentiated non-finite counter.** Acknowledged. A `reason` label (or separate counters for NaN-ce / NaN-boost / NaN-clip / negative-clip / ±inf-boost) is genuinely useful in production triage. Filing as a follow-up V-ticket alongside the V17 observability work — they share the same "operator-facing reflection on retrieval composition" surface.

**L2 — `float | object` validator return type.** Acknowledged. `float` is more precise. The `mode='before'` contract means Pydantic re-validates downstream, so the runtime behavior is identical, but the annotation could communicate the always-numeric outcome more cleanly. Deferred for the same follow-up commit; not patching in this round to avoid burning a fresh review cycle on a one-line annotation tightening.

**L3 — serializer `None` produces non-roundtrippable JSON.** Acknowledged. The `null` output is intentionally defensive (better than `Infinity`/`NaN` non-RFC tokens) but a config file with `null` would fail load. The only path that produces this is *post-construction attribute mutation* with a non-finite value, which itself violates the `ge=0.0` contract. A one-line code comment noting "not a valid deserialization target" is the right fix; will fold in with the L1 follow-up.

**L4 — Prometheus globals across test boundaries.** Acknowledged. The test file isn't run under `pytest-xdist` in this repo, but a `pytest.mark.unit` or module-level comment about the shared-state assumption is reasonable hygiene. Deferred.

### Closing self-assessment

**99%**, unchanged. Gap: -1% for the three LOW findings outside V19's diff scope that V19 surfaced (pre-existing stale anchors in eval docs and test docstrings — not patched here). All Hermes round-1-through-4 MEDs addressed in commits 77c26682 / 29874625 / cef24043; the round-4 LOWs go into a follow-up V-ticket.

PR ready for human review.

PR #169, opened 2026-05-14
feat(reflection): split per-entity advisory lock from LLM phases (V18)

JasperHG90:
## Summary

Stops holding a Postgres advisory transaction lock across 30-60 s of DSPy LLM I/O per entity reflection. Replaces `pg_try_advisory_xact_lock(hashtext('reflect:{entity_id}'))` with:

- **Process-local `asyncio.Lock` per entity** (via `WeakValueDictionary` registry) — intra-worker serialization without holding a DB connection.
- **Compare-and-Set UPDATE on `MentalModel.version`** for the Phase 5 write — cross-worker write safety without a long-running transaction.
- **Per-entity short DB sessions** — Phases 0-4 + 6 run with no open transaction; Phase 5 opens a fresh session for the CAS write and closes immediately.

Eliminates connection-pool exhaustion under concurrent reflection load and the long-running MVCC snapshots that were blocking `VACUUM`.

## Surfaces

Reflection is a background worker — no direct agent surface. However, ConfidentResolutionAbandonedError raised on CAS-loss is propagated end-to-end:

- HTTP `POST /reflection/summarize_node` → **503** with `{retry_after_seconds, hint}` envelope (`retry_after_seconds` derived from `summarize_node_rate_limit.per_entity_per_seconds` config; pre-V18 leaked as 500).
- MCP `memex_summarize_node` → 503 envelope translated to MCP-friendly response with `retry_after_seconds` + `hint` fields; tool description updated.
- Hermes `memex_summarize_node` → mirror surface; tool schema description updated.
- Common client `ReflectionAbandoned` exception class added; 503 parser surfaces structured fields.
- `Locks.reconsolidate_entity` response gains `abandoned: bool` (default `False`).

## Files

- `packages/core/src/memex_core/memory/reflect/reflection.py` — Phase 5 CAS, per-entity asyncio.Lock acquisition, `_entity_session()` helper, Phase 6 expire-instead-of-expunge
- `packages/core/src/memex_core/memory/reflect/entity_locks.py` — new asyncio-lock registry helper
- `packages/core/src/memex_core/memory/reflect/exceptions.py` — new `ReflectionAbandonedError`
- `packages/core/src/memex_core/memory/reflect/queue_service.py` — new `mark_abandoned` (PENDING without retry_count increment, preserves prior `last_error`)
- `packages/core/src/memex_core/services/reflection.py` — `reflect_batch_detailed` returns `tuple[list[ReflectionResult], list[UUID]]` (race-safe vs the singleton field it replaces); `reflect_batch` is a thin wrapper
- `packages/core/src/memex_core/services/locks.py`, `services/notes.py` (background flow) — call-site updates
- `packages/core/src/memex_core/server/reflection.py`, `server/memories.py` — 503 envelope
- `packages/core/src/memex_core/scheduler.py` — narrow OSError / RuntimeError / IntegrityError / OperationalError catch (down from bare `Exception`)
- `packages/core/src/memex_core/metrics.py` — `REFLECTION_CAS_ABANDONS_TOTAL` counter; corrected V19 composite-boost docstrings
- `packages/common/src/memex_common/client.py` — `ReflectionAbandoned` exception
- `packages/mcp/src/memex_mcp/server.py`, `_summarize_descriptions.py` — envelope translation + description
- `packages/hermes-plugin/src/memex_hermes_plugin/memex/tools.py` — Hermes envelope + schema description (also fixes pre-existing 1000→100 unit cap drift)
- Tests:
  - `tests/test_e2e_f5_summarize_node_http.py` — 503 envelope assertion
  - `packages/mcp/tests/test_summarize_node_envelopes.py` (new) — MCP envelope contract
  - `packages/hermes-plugin/tests/test_tools.py` — Hermes envelope tests
  - `packages/core/tests/integration/test_int_reflection_lock_concurrency.py` (new) — CAS regression guards + retry_count invariance under contention
- `docs/explanation/reflection-and-mental-models.md` — surface-adapter behavior on abandons section

## Pre-PR gates

- Self-assessment: 99%
- Adversarial review loop: 10 rounds (each round patched + a fresh reviewer per round); final round returned 0 CRIT / HIGH / MED.

## Test plan

- [ ] `uv run pytest packages/core/tests/integration/test_int_reflection_lock_concurrency.py -v` → CAS regression + retry_count invariance
- [ ] `uv run pytest packages/mcp/tests/test_summarize_node_envelopes.py packages/hermes-plugin/tests/test_tools.py -v` → envelope round-trip on both agent surfaces
- [ ] Two reflection workers on different entities + mocked DSPy with `asyncio.sleep(N)` → both complete in ~N seconds (no serialization across entities)
- [ ] `pg_stat_activity` shows no transaction older than 5s during steady-state reflection
- [ ] E7 baseline ranking gate green (no retrieval-path regression)


JasperHG90:
## Round-10 patch + MED-3 rebuttal

Pushed `e767f985` addressing round-1 review:

| ID | Severity | Status |
|----|----------|--------|
| H1 | HIGH | Fixed — `last_abandoned_entity_ids` instance attribute removed; `reflect_batch` now returns `(list[MentalModel], list[UUID])` and `_process_entity_reflection` threads the list through as a parameter. Concurrent invocations no longer race. |
| H2 | HIGH | Fixed — `assert self.lm is not None` replaced with explicit `if self.lm is None: raise RuntimeError(...)` (assert silently optimized out under `python -O`). |
| M2 | MED | Fixed — added full type annotations on `queue_service.get_next_batch` (`vault_id`, `vault_ids`). |
| M4 | MED | Fixed — added comment on `entity_locks` module-level lock explaining Python 3.12 `asyncio.Lock` semantics that make the construction safe. |
| M5 | MED | Fixed — `test_int_reflection_lock_concurrency.py::test_reflect_batch_tracks_cas_abandons_separately_from_success` unpacks the tuple directly; no more reliance on now-removed instance state. |
| M3 | MED | **Rebuttal — false positive (see below).** |

### M3 rebuttal — `ReflectionAbandoned` is reachable from the client

The reviewer claimed `ReflectionAbandoned` is defined but never raised by the client. This is incorrect.

`packages/common/src/memex_common/client.py:1049-1088` (`RemoteMemexAPI.summarize_node`) maps the HTTP envelope back to the exception:

- HTTP 503 with `code=="reflection_abandoned"` → `raise ReflectionAbandonedError(...)`
- The server emits this envelope from `services/reflection.py::reflect()` when the engine returns no models but the entity is in the abandoned set.

The exception is raised from the client wrapper. Catching it in agent-facing code (e.g. MCP tool handlers, Hermes wrapper) is the documented integration point; the client surface is what makes it reachable.

Tuple-unpack changes rippled through every caller in core/tests; full reflection suite (1,476 tests) is green. Re-running Hermes review.

JasperHG90:
## Round-11 patch + rebuttals

Pushed addressing the round-2 review.

| ID | Severity | Status |
|----|----------|--------|
| C1 | CRITICAL | Fixed — `reflect_batch` now returns `(success_models, abandoned_ids, failed_ids)`. The three buckets are disjoint and routed through `complete_reflection` / `mark_abandoned` / `mark_failed` respectively in both `services/reflection.reflect_batch_detailed` and `memory/engine.process_reflection_queue`. Regression test: `test_reflect_batch_routes_real_exceptions_to_failed_bucket`. |
| H2 | HIGH | **Rebuttal — false positive (see below).** |
| H3 | HIGH | Fixed (folded into C1) — `asyncio.gather(..., return_exceptions=True)` so a `BaseException` cannot tear down sibling coroutines mid-flight; the inner `except Exception` already trapped normal exceptions and now appends to `failed_entity_ids` explicitly. |
| M5 | MED | Fixed — `reflect_batch` dedupes by `entity_id` at entry. Regression test: `test_reflect_batch_dedupes_duplicate_entity_id_requests`. |
| M6 | MED | Fixed — comment added explaining the Phase 5 expunge vs Phase 6 expire ownership split. |
| M7 | MED | **Rebuttal — false positive (see below).** |
| M8 | MED | Fixed — `_ensure_queue_items` now tolerates the documented row shapes (bare UUID, tuple-wrap, `.entity_id`, `.id`, stringified) with a swallow-on-coercion-failure fallback and a comment explaining why. |
| L10 | LOW | Fixed — replaced all four remaining `assert self.lm is not None` in Phase 1/3/4/5 with explicit `if self.lm is None: raise RuntimeError(...)`. |
| L9, L11 | LOW | Deferred (defensive logging conventions / config coupling — not regressions, separate-PR scope). |
| L12 | LOW | **Rebuttal — false positive (see below).** |

### H2 rebuttal — WeakValueDictionary GC race is not reachable

The reviewer claims that between `_entity_locks.get(entity_id)` returning a lock and the caller using it in `async with`, the WeakValueDictionary can drop the entry because "no strong reference exists on the event loop."

This is not how CPython references work. Between those two points the lock is bound to a local variable in the calling frame (`lock = await get_entity_lock(...)`). Local variables in active stack frames are strong references; suspended coroutines keep their frames alive (with all locals) until they resume and either return or raise. So the registry entry is always pinned by *every* current holder/waiter for as long as they're in scope. The failure mode the reviewer describes would require garbage-collecting a live Python frame, which CPython never does.

If two coroutines call `get_entity_lock(eid)` concurrently, both walk the same `_entity_locks.get()` → both get the same `L` (the first creates it under `_registry_lock`, the second observes it after the lock is released). They serialize on `async with L:` exactly as intended. The "different locks for the same entity" scenario the reviewer sketches requires the registry entry to be dropped while a caller is using it — which the strong-reference rule above forbids.

### M7 rebuttal — `ReflectionAbandoned` docstring is correctly triple-quoted

`packages/common/src/memex_common/client.py:75-86` is a single triple-quoted docstring starting `"""Raised when...` on L76 and closing `"""` on L86. The body contains an embedded `"re-read..."` quotation — a single `"` inside a triple-quoted string is valid Python and does not need to be escaped or matched. There is no stray opening quote.

### L12 rebuttal — `entity` is a function parameter, not undefined

`test_int_reflection_lock_concurrency.py::test_two_workers_different_entities_run_in_parallel` defines `async def run_reflection(entity, model)` (a local helper). The `entity.id` reference inside resolves to the helper's parameter, which is bound to `entity_a` and `entity_b` at the two `gather()` call sites. No `NameError` is possible.

---

Round-11 commit: 1 file/13 changed, 1,478 reflection-related tests green. Re-triggering Hermes.

PR #170, opened 2026-05-15
feat(core): close deprioritization leak via observation refresh worker

JasperHG90:
## Summary

Closes the deprioritization leak via mental-model observations: when a memory unit is deprio'd, observations citing it are now re-synthesized on the surviving evidence (or dropped if no evidence survives) by a new refresh-observation worker, instead of being silently re-served as virtual units. Passing an observation UUID to `memex_memory_deprioritize` returns 400 with `source_memory_units` redirecting the caller to the underlying MU IDs.

## Scope

**Schema** (migrations 043 + 044)
- `reflection_queue`: `task_type` / `observation_id` / `priority_lane` / `source_unit_id` + 3 partial indices (claim ordering, refresh dedupe, restore upsert target). Partial UNIQUE on `(entity, vault, observation_id) WHERE status IN ('pending','processing')` dedupes refresh enqueues. CheckConstraint pins `task_type` domain.
- `mental_models`: GIN(`observations` `jsonb_path_ops`) for vault-scoped `@>` scans during deprio.

**Observation identity**
- Phase 4 reconstruction preserves observation UUIDs via structured `ObservationProvenance` (lowest existing index wins on merge/kept). Per-entry validation falls back on malformed indices with labeled metrics; UUID collision tracker prevents two outputs sharing one id.
- Retrieval `_convert_mm_to_units` derives `virtual_id` from `observation.id` — agent-visible identity is now stable across reflection cycles.

**Refresh worker** (`_refresh_observation`)
- Three-phase shape: read tx → LLM call (no DB session held) → write tx with CAS UPDATE on `version`. Phase C re-validates live evidence against current state to close the race where a concurrent deprio during Phase B silently dedupes its refresh enqueue against the in-flight PROCESSING row.
- Drop guardrail: LLM `should_drop=True` overridden when `len(live_ids) >= min_evidence_for_obs_retention` (default 2).
- Cross-vault evidence filtered defensively.

**Service / route**
- `_flip_deprioritized`: vault-scoped JSONB scan with `FOR UPDATE OF mental_models`; bulk-insert refresh tasks on commit; rollback rolls both back. FSFM auto-band defers per-unit refresh and flushes once via `LATERAL unnest`. `batch_set_unit_deprioritized` scoped by `vault_id`.
- HTTP route + `_handle_error` catch `ObservationReadOnlyError` first and surface a **dict** detail (not a flattened string) containing `source_memory_units`.

**Agent surface**
- `MEMEX_MEMORY_DEPRIORITIZE_DESC` and `agent_surface.VIRTUAL_UNIT` / `CRITICAL_HEADER` / `CRITICAL_FOOTER` state the HTTP 400 contract with `source_memory_units`. SSOT pulled across MCP + Hermes via re-export of `tool_descriptions`. AGENTS.md tier-1b cap reconciled to 6,500 chars.

**Scheduler**
- Partition claimed batch by `task_type`; dispatch refresh BEFORE reflect within a batch; per-item exception isolation; `AdvisoryLockTakenError` reclaims via jittered backoff WITHOUT bumping `retry_count`. Vault-partitioned reconcile pass repairs missed enqueues using a stable cursor over sorted vault IDs (gated behind `reconcile_historical_deprios_on_boot`).

## Adversarial review

Three full review rounds (concurrency / contracts / schema) run to clean. Notable finds:

- **Round 1** (5 CRITICAL, 8 HIGH, 13 MEDIUM): ImportError on every refresh dispatch; `enqueue_priority_reflect` double-commit; transient LLM `None` silently dropping queue rows; advisory-lock asymmetry with Phase 5; vault scope missing on `live_ids`; partial-UNIQUE arbiter inference inconsistency; …
- **Round 2** (6 HIGH, 6 MEDIUM): tx span across LLM call (Phase A/B/C split — no DB session held during LLM); 3 enqueue counter overcount sites (use `result.rowcount` after `on_conflict_do_nothing`; RETURNING `xmax = 0` for upserts); idempotent-restore audit emission restored (FSFM cooldown dependency); GIN + 3 partial indices added to `__table_args__` to prevent autogenerate drift; …
- **Round 3** (1 CRITICAL, 2 HIGH, 3 MEDIUM): **race regression** — concurrent deprio mid-Phase-B silently deduped against in-flight PROCESSING row; closed by Phase C re-validation that raises `AdvisoryLockTakenError` (reclaim, no retry bump) when `current_live_ids != phase_a_live_ids`. Also: `(xmax = 0)` fragility hardened to `sql_text('(xmax = 0) AS was_insert')`; service session no longer held across LLM call; sibling refresh-task query filtered to in-flight rows.
- **Round 4**: verified the race-regression fix closes the leak; one LOW (exception message wording) fixed.

## Test plan

- [x] 1771 unit tests pass on V21-touched paths (`packages/core/tests/unit/memory/reflect/`, `packages/core/tests/unit/services/`, server, retrieval, common). 1 pre-existing F45 metrics flake confirmed not V21-induced.
- [x] `just prek` clean (ruff + format + mypy).
- [ ] Integration tests (testcontainers Postgres) for partial UNIQUE arbiter inference, FOR UPDATE serialization, GIN plan, scheduler dispatch — pending in CI.
- [ ] `feedback_deprioritize_obsolete` eval scenario replay — pending in CI.
- [ ] Manual MCP smoke: deprio an observation UUID → expect 400 with `detail.source_memory_units`; re-issue against an MU ID → expect 200 and refresh task ticked.

## Files

32 files changed, +2595/−57 LOC.


JasperHG90:
## Patch round-1 — Hermes findings

Pushed `9db4b074` addressing the review.

**HIGH 1** — `postgresql_ops` sort-direction misuse on `idx_reflection_queue_lane_priority`. Fixed; rendered DDL now matches migration 043 byte-for-byte (`priority_lane DESC, priority_score DESC, last_queued_at`). Verified locally with `CreateIndex(ix).compile(postgresql.dialect())`.

**MEDIUM 2** — `AdvisoryLockTakenError` semantically conflated. Added two subclasses inheriting from it: `RefreshStaleReadError` (Phase A vs Phase C live-evidence diverged) and `RefreshCASAbandonedError` (CAS UPDATE rowcount=0). Scheduler `except AdvisoryLockTakenError` keeps working — the change is self-documentation at the raise site.

**MEDIUM 3** — `ComparePhaseOutput.changes_summary` → `provenance` schema change for DSPy cache busts: acknowledged. The reflection LLM uses DSPy `Predict` with structured Pydantic output — there is no compiled-program cache I'm aware of in this codebase (`grep dspy.load\|compiled_state\|teleprompt`). If a compiled-program cache is added in the future it should be invalidated on signature schema changes. Will flag in the design-doc / followup. Not patching code here.

**MEDIUM 4** — `_find_source_mus_for_observation` multi-match: now logs `logger.error` with the collision count before returning the first match. UUID4 collision is an invariant violation; loud signal is correct.

**MEDIUM 5** — `refresh_obs_priority_lane` default-True starvation risk: docstring expanded with operational threshold (`queue depth >100`) and remediation (flip False to let refresh compete on score).

**LOW 6** — `_drop_observation_in_place` test-only: kept in production module. Moving to a test fixture would cross package layout boundaries and create churn for a 9-line helper. Docstring already states "retained for test scaffolding"; CAS pattern is the production contract.

**LOW 7** — README test badge `6,557 → 6,113`. Investigated: `uv run pytest --collect-only -q` returns `6113 tests collected, 3 errors` on both the V21 branch AND on the `release/tier-a-cognitive-memory` base (verified via `git stash` round-trip). The 3 collection errors are pre-existing (`memex_hermes_plugin` import + 2 MCP test imports) and not V21-induced. The previous badge value of 6,557 must have been captured on a branch with patches that fixed those imports. No V21-introduced test deletions.

**LOW 8** — `_reconcile_one_vault_slice` `getattr` chains: kept as-is for V21. Worth a follow-up to migrate config access to typed accessors codebase-wide.

**LOW 9** — `reflect_input_only` double-negative naming: kept; the SQLAlchemy convention is positive-flag-with-False-default, and the docstring is explicit.

**LOW 10** — `batch_set_unit_deprioritized` skips flush when `vault_id is None`: now logs a `WARNING` with unit count so operators see the legacy-caller path being taken.

`just prek` clean. 407 V21-touched unit tests + 1,486 broader unit tests pass.

PR #171, opened 2026-05-15
Retrieval-stability suite — memory + note rerank regression gate

JasperHG90:
## Summary

Refits the retrieval-stability regression gate as a real suite under
the eval-suite framework (`packages/eval/src/memex_eval/suites/retrieval_stability/`).
Replaces the previous parallel-infrastructure approach (standalone harness +
`tests/test_e2e_*`) which sat outside the framework.

- **102 scenarios**: 51 unique queries × 2 search modes (`memory`, `note`).
- **Suite-private extensions** register via decorator at import time;
  framework core (`packages/eval/src/memex_eval/suite/*`) untouched.
- **Both memory-rerank and note-rerank paths** gated — the memory path
  exercises the log-additive bounded-boost composition; the note path
  exercises the independent `CE → sigmoid → RRF → cosine` pipeline.
- **Note-key anchoring** (filename stems) — survives the cross-machine
  cache miss that would break unit-id anchoring silently.
- **Env-gated capture mode**: `MEMEX_EVAL_CAPTURE_BASELINES=1 memex-eval
  suite run retrieval_stability` writes per-scenario baselines with a
  meta block (`schema_version`, `top_k`, `search_type`); verify mode
  reads the meta back and refuses to compare against a stale baseline.

The suite ships without captured baselines — the README documents the
operator workflow: capture-then-verify on first run after a fresh
clone or after an intentional retrieval change.

The framework-extension example (`seed_paragraphs_from_sources` setup
action) is shipped per the directive that the framework should allow
direct-DB seeding. The suite itself uses normal ingest; the handler is
exercised against a testcontainer Postgres so it stays warm-tested.

`.claude/rules/eval-suites.md` is added so future agents adding or
extending eval suites don't drift into parallel infrastructure.

## Test plan

- [x] 16 RBO math unit tests pass (`packages/eval/tests/test_ranking_stability.py`)
- [x] 24 outcome unit tests pass (memory + note search, capture/verify, meta-mismatch guards, defensive meta-block guard)
- [x] 12 setup-action integration tests against testcontainer Postgres pass (deterministic UUIDv5, BOM/CRLF, idempotency, missing-params, registration)
- [x] Suite discovers cleanly on the CLI: `memex-eval suite list` shows `retrieval_stability` with 102 scenarios
- [x] No framework files modified: `git log release/tier-a-cognitive-memory..HEAD -- packages/eval/src/memex_eval/suite/` returns empty
- [x] No ticket IDs in shipped surfaces (verified by grep across code, docstrings, comments, error messages)
- [x] Four rounds of adversarial review; loop exited clean (zero CRIT/HIGH/MED on round 4)
- [ ] Capture step end-to-end against a live Memex server — not run on this branch; operator/CI captures on first local run per the documented workflow


JasperHG90:
Bot review on `6ad8e16e` triaged. Actionable items closed in `d2af980c`:

- **HIGH #3** (env-mutation invisibility): emit one explicit log line at module-import time when `MEMEX_EVAL_CAPTURE_BASELINES` is set. Per-call env reads remain in `score()` so tests using `monkeypatch.setenv(...)` still observe their flips.
- **MED #5** (frontmatter regex stripping body-level `---`): documented the source-file convention explicitly in the regex comment.

Deferred with rationale:

- **CRIT #1** (`fcntl.flock` Windows incompatibility, `snapshot_cache.py:233`): Pre-existing framework code, NOT introduced by this PR. `git diff release/tier-a-cognitive-memory..HEAD -- packages/eval/src/memex_eval/suite/snapshot_cache.py` shows this PR only added `compute_sources_hash`; `_slot_lock` predates the branch. The project is Linux-server-first and CI runs on Linux runners. Bonafide Windows compatibility for the eval snapshot cache is a separate decision and belongs in its own PR — out of scope here under the project rule "only change eval code when fixing eval scenarios."
- **HIGH #2** (sync `score()` event-loop blocking, `_outcomes.py:131`): The framework's `ExpectedOutcomeBase.score()` is sync by design — every existing outcome (`KeywordsPresent`, `GoldUnitIds`, `RankingOrder`, `UsefulAtK`, ~20 more in `base.py`) returns synchronously. Making this single outcome async would either (a) break the framework's score-dispatch contract or (b) require an `asyncio.to_thread` wrapper at every other outcome's call site. Capture mode is operator-triggered (`MEMEX_EVAL_CAPTURE_BASELINES=1`), runs rarely, and writes ~300 bytes per scenario; the event-loop impact is negligible. Framework-wide refactor to async score() is a separate decision.
- **MED #4** (RBO O(n²) past `short_len`): Pre-existing code in `ranking_stability.py`; not touched by this PR. The complexity-warning comment is already in the docstring (negligible at k=10–20 which is the suite's range). If extended to longer rankings, a length guard belongs in its own PR.
- **MED #6** (`config_pins` dict comparison): Bot acknowledged the `!=` is correct ("This is correct"). Not actionable.
- **MED #7** (`shipped_snapshot_path` validation): Bot acknowledged the trusted-author model is intentional ("same as test fixtures"). The framework already requires trusted authors to register suites; a path-confinement check would not raise the trust bar meaningfully. Not actionable.

JasperHG90:
Bot review on \`6ad8e16e\` (rerun, 5m32s) triaged. 1 HIGH + 5 MEDIUM addressed in \`d838d253\`:

- **HIGH** (null-valued required meta keys bypass value-comparison guards): added a null-value validator alongside the existing missing-key check; rejects null on any of \`top_k\` / \`search_type\` / \`schema_version\` / \`config_pins\` with the same recapture hint. Parametrized test added for the four keys.
- **MED** (\`_slugify\` empty-string edge case): falls back to \`'_unknown'\` instead of producing double-underscore scenario ids.
- **MED** (\`refresh-snapshot\` restore on partial copytree): pre-restore cleanup widened from \`is_dir()\` to \`exists()\` so a stray file, partial dir, or symlink at the shipped path is removed before the backup is moved back into place.
- **MED** (\`--reuse-vault\` silently bypasses shipped snapshot): added a runner info log so the operator sees the bypass before the verify run fails.
- **MED** (DSN password fallback in seed handler): log a warning when the DSN has no password and we fall back to the testcontainer default \`'postgres'\`. Also moved \`metastore.connect()\` inside the try block per the bot's nearby finding so a connect-time failure routes through the close() cleanup.

Deferred LOW findings with rationale:

- **LOW** (RBO docstring lacks dedup semantics note): \`ranking_stability.py\` is pre-existing code; the math is correct per the standard RBO definition. Docstring polish can be a follow-up.
- **LOW** (\`_write_baseline\` temp-file race): each scenario has its own \`baseline_path\` (per-scenario file under \`baselines/\`), so two scenarios sharing a \`.json.tmp\` path is not reachable without dynamic-scenario construction that doesn't exist today.
- **LOW** (pg18-trixie container tag): \`pg18\` is the testcontainer tag for PostgreSQL 18 + pgvector; the tag is currently published. If the upstream image is renamed, the integration test fails fast at container-pull, not silently.
- **LOW** (no \`metric_keys()\` assertion): the framework's runner asserts \`metric_keys()\` indirectly via the MLflow logger; adding a dedicated unit test is doc-quality, not coverage.

Deferred MEDIUM with rationale:

- **MED** (\`ast.walk\` order is unspecified): \`ast.walk\` is documented as breadth-first traversal in CPython; this is part of the public API contract, not an implementation detail. Inserting an "is this two-tier" guard would catch a corpus problem that doesn't exist today.

Bot CRITICAL #1 from the prior review (fcntl Windows incompatibility) remains deferred — pre-existing framework code, out of scope per the project rule (rebuttal in earlier comment).

JasperHG90:
Bot review on \`d838d253\` triaged.

Closed in \`927bdbd6\`:
- **HIGH** (DSN password fallback): hard-fail unless \`MEMEX_EVAL_ALLOW_DEFAULT_POSTGRES_PASSWORD\` is explicitly set; bare warning escalated to error-level when the opt-in path fires. Testcontainer integration tests pass because \`PostgresContainer\` injects a real password into the DSN — the branch never fires there.

Deferred with rationale (pre-existing framework code, NOT introduced by this PR, per the project's "only change eval code when fixing eval scenarios" rule):
- **HIGH** (\`runner.py:372\` \`import base64\` inside per-note loop): blame shows the line is from commit \`47a77d426\` (2026-05-08, before this branch). Hoist-the-import is a follow-up.
- **MED** (\`runner.py:435\` idempotent-skip uses raw \`str()\` instead of \`_canonical_uuid()\`): pre-existing framework code.
- **MED** (\`runner.py:209\` O(n²) list-membership check): pre-existing framework code; the bot acknowledged the comment-pinned back-compat reason.
- **MED** (\`snapshot_cache.py:266\` leaked \`.old-\` dirs on crash): pre-existing framework code; design tradeoff documented in the snapshot-cache module.
- **MED** (\`decorator.py:217\` asyncio.run fallback): pre-existing framework code; test-only code path with explicit module-level guard.

Deferred LOW findings (\`_outcomes.py:179\` redundant \`list()\` on already-list slice, \`agents.py:154\` registry-name pattern inconsistency, pg18-trixie image pinning, base.py \`_aggregate_text\` field-fallback comment) — cosmetic; will fold into a follow-up if any second consumer requires the same outcome.

Atomic-write Windows cross-device concern (MED in \`_outcomes.py:316-322\`): Memex is Linux-only by design; \`pathlib.Path.replace\` semantics on Linux are atomic on the same filesystem, which baselines/ always is. Not in scope.

JasperHG90:
Bot review on \`927bdbd6\` triaged (0 CRIT/HIGH, 4 MED, 6 LOW).

Closed in \`1b0bab62\`:
- **MED** (capture writes to source tree): documented in README that \`MEMEX_EVAL_CAPTURE_BASELINES\` requires editable install.
- **MED** (\`_scrape_queries_from_suite\` silent omission): added \`TestScrapedQueryFloor\` with per-corpus floor counts (acme_corp=36, ai_research_lab=7, project_nexus=8). Source-suite refactor that drops a literal \`query=\` kwarg now fails at CI.

Deferred (pre-existing framework code, NOT introduced by this PR):
- **MED** (lint findings 500-row cap, \`agents.py:314\`).
- **MED** (\`--notes-file\` path validation, \`cli.py:771\` in suite_run — \`refresh-snapshot\` does not have a notes-file flag).

Deferred LOW with rationale:
- (paragraph-min-chars not ClassVar, slugify aliasing, bare except in DirectApiBackend, asyncio.run in CustomEvaluate, fcntl Windows, metadata in-place mutation): all pre-existing framework conventions or design choices acknowledged in code comments.

JasperHG90:
Bot review on \`1b0bab62\` triaged. **No code changes needed this round** — all findings are either non-physical, mitigated by existing tests, or contradict the bot's earlier recommendation.

- **HIGH** (DSN password fallback "even with env-gate", \`_setup_actions.py:241-269\`): the design in \`927bdbd6\` is exactly what the previous review (#26) recommended — "elevating the log.warning to log.error **or** adding a hard-fail mode." This round implements **both**: hard-fail by default (raises with a clear pointer-to-env-var); opt-in escape hatch with error-level logging when \`MEMEX_EVAL_ALLOW_DEFAULT_POSTGRES_PASSWORD\` is set. Removing the opt-in entirely would lock the handler when a testcontainer DSN ever strips the password (older testcontainer versions, custom CI shims). The escape hatch is gated, loud, and documented. Marking as resolved.
- **MED** (AST scraper misses \`@suite.scenario\` decorators / \`register_class\` calls): no source suite under \`packages/eval/src/memex_eval/suites/\` uses decorator-style scenario registration — verified via grep. The round-8 \`TestScrapedQueryFloor\` guard catches any future refactor that drops queries via this path (per-corpus floors: 36/7/8). Mitigation is in place.
- **MED** (\`_aggregate_unit_ids\` falsy-id short-circuit with integer \`0\`): \`MemoryUnitDTO.id: UUID\` is Pydantic-validated; an integer value cannot exist at this layer. Even the all-zero UUID is truthy in Python (\`bool(UUID(int=0)) == True\`), so the \`or\` short-circuit is correct for all reachable types. Theoretical concern only.

Deferred LOWs (atomic-write EXDEV check, \`_BASE_SCENARIO_MUTABLE_LIST_FIELDS\` fragility): cosmetic / pre-existing framework architecture.

PR remains green on substantive findings; CI gates (path-detection + Hermes review) both pass.

JasperHG90:
Round-10 bot HIGHs closed in \`bbb462b1\`:

- **HIGH #1** (namespace UUID pinning): added \`test_namespace_uuid_is_pinned_to_committed_value\`. A refactor that re-randomizes the namespace now fails at CI.
- **HIGH #2** (frontmatter regex \`\\n?\`): expanded inline comment to document both branches.
- **HIGH #3** (capture-mode parallel-writer race): added per-path advisory \`flock\` + PID-suffixed tempfile names. Two concurrent capture-mode writes on the same baseline path now serialise. README documents the single-writer contract.
- **HIGH #4** (positional \`query\` args invisible to scraper): scraper now emits a \`logger.warning\` when register/scenario calls have positional args. AST scan confirms no current source suite uses positional args; the round-8 floor-count test is the second safety net.

All 59 retrieval_stability tests pass. PR ready for human review.

JasperHG90:
Round-11 bot MEDs closed in \`9a42a951\`:

- **MED #1** (dead \`is not None\` guards): removed redundant null-check short-circuits on the value-comparison guards. The round-9 null-meta-key validator catches null-valued keys earlier; the secondary guards were dead on the null path. Switched from \`.get()\` to direct \`[key]\` lookup since presence is also guaranteed by the missing-key validator.
- **MED #2** (\`_slugify\` empty-slug collision): two empty-slug queries in the same corpus would both fall back to \`'_unknown'\` and silently collide on the scenario_id. Appended an 8-char SHA-256 prefix of the original text to disambiguate.

Three LOWs deferred (flock-unlock-before-close micro-window: benign; snapshot_cache publish recovery: pre-existing framework; \`vault_id: Any\` typing: pre-existing convention).

Down from 4 HIGH (round 10) → 0 HIGH / 2 MED (round 11) → expected 0 HIGH / 0 MED / 3 LOW after this commit. Ready for human review.

JasperHG90:
Round-12 bot MEDs closed in \`19d8ccaa\`:

- **MED** (duplicate \`os\` import with \`_os\` alias in \`_outcomes.py\`): \`os\` is already imported at module top; the local \`import os as _os\` inside \`_write_baseline\` was unnecessary and confusing. Switched to \`os\` directly; kept the \`_fcntl\` alias since the import-fallback pattern needs a guarded name.
- **MED** (\`_stable_unit_id\` accepts negative \`paragraph_index\`): added explicit \`paragraph_index < 0\` guard + matching test. Internal callers use \`enumerate(...)\` so this is always ≥ 0, but the function is part of the public framework-extension surface.

Third MED (TemporalOrdering iteration ordering in \`base.py:669\`) is pre-existing framework code, not introduced by this PR — out of scope.

Six LOWs deferred with rationale in the commit message. Trajectory: 4 HIGH (round 10) → 0 HIGH / 2 MED (round 11) → 0 HIGH / 2 MED (round 12; one pre-existing) → expected 0 HIGH / ≤1 MED (round 13).

JasperHG90:
Round-13 bot LOWs closed in \`0e3e14f0\`:

- **LOW** (\`test_atomic_write_no_tmp_file_left\` vacuously true): the round-10 PID-suffixed tempfile rename made the literal \`s1.json.tmp\` filename unreachable; the assertion checked for a file that was never created. Replaced with a glob over \`s1.json.tmp*\` so the test actually exercises the cleanup contract.
- **LOW** (\`_scrape_queries_from_suite\` match-scope): documented the permissive matcher in the docstring (matches any \`register\`/\`scenario\` call, not just \`suite.<x>(...)\`). No current source suite has a non-suite \`register\` call; floor-count test guards count regressions.

**Iterate-until-clean exit**: round 13 found 0 CRIT/HIGH/MED. The recursive review cycle is clean.

Trajectory:
- Round 10: 4 HIGH
- Round 11: 0 HIGH / 2 MED
- Round 12: 0 HIGH / 2 MED (one pre-existing)
- Round 13: 0 HIGH / 0 MED / 2 LOW
- (After this commit: round 14 expected clean.)

PR is ready for human review/merge.

PR #173, opened 2026-05-18
refactor(eval): scenario data/registration split + read_scenario_specs primitive

JasperHG90:
## Summary

Replaces the AST scraper in ``retrieval_stability`` with a framework primitive — ``memex_eval.suite.read_scenario_specs(name)`` — and migrates the three upstream corpora (acme_corp, ai_research_lab, project_nexus) to a data/registration split so that downstream consumers can enumerate scenarios without paying the cost of full suite construction.

## Why

``retrieval_stability`` needs the queries from the three source corpora but MUST NOT import them — importing the parent package would fire all setup-action registrations, eagerly import ``memex_core``, and contaminate the runner's state. The prior workaround was an AST scraper that walked ``__init__.py`` looking for ``query=`` literals. That is fragile (query rename = silent drop), architecturally backwards (data extraction via syntax parsing), and breaks if a query value isn't a string literal.

The right shape is data/behavior separation: scenarios are pure data (registration kwargs), registered into a Suite at import time, and readable independently.

## What changed

**Framework**

- ``memex_eval.suite.read_scenario_specs(name) -> list[dict]`` — loads a suite's ``scenarios.py`` via ``spec_from_file_location``, bypassing ``__init__.py`` entirely. Returns the raw list of register-kwargs dicts.
- Contract: ``scenarios.py`` MUST use absolute imports only (``from memex_eval.suite import X``).
- Re-exported from ``memex_eval.suite``.

**Migrated suites** — each now ships:

- ``scenarios.py`` — module-scope ``SCENARIO_SPECS: list[dict]``, one entry per ``suite.register(...)`` call.
- ``__init__.py`` — thin wrapper that builds the Suite, imports ``SCENARIO_SPECS``, and replays the registrations.

Counts: acme_corp (40), ai_research_lab (9), project_nexus (11).

**retrieval_stability** — replaces ``_scrape_queries_from_suite`` (AST walker) with ``_queries_from_source_suite`` (calls ``read_scenario_specs``, projects the ``query`` field). Builds 100 scenarios without importing any source suite's parent package.

**Tests**

- ``TestReadScenarioSpecsIsSideEffectFree`` — parametrized over all three source corpora; asserts the parent package is absent from ``sys.modules`` after the read.
- ``TestSourceSuiteQueryFloor`` (renamed from ``TestScrapedQueryFloor``) — floor assertions retarget the new helper.

**Snapshot manifest** — alembic_head 044 → 045 to track the procedure-outcomes drop (data-equivalent: snapshot dump had 0 rows for that table).

## Test plan

- [x] ``uv run python -c \"import memex_eval.suites.acme_corp; ...\"`` — all four suites build cleanly (40 / 9 / 11 / 100 scenarios).
- [x] ``uv run ruff check`` — passes.
- [x] Pre-commit hooks pass (ruff, ruff-format, mypy, test-count badge).
- [x] 211 targeted tests pass (retrieval_stability + outcomes + extensibility).
- [ ] CI green on push.


JasperHG90:
Closing — wrong approach; abandoning the branch.

PR #183, opened 2026-05-28
refactor: procedures live UNDER a scope namespace, not bare

JasperHG90:
## Summary

Procedures are no longer a top-level KV namespace. They live UNDER an existing scope namespace as `<scope>:procedure:<verb>:<context-tag>`:

- `global:procedure:<verb>:<context-tag>` — default global procedure
- `project:<id>:procedure:<verb>:<context-tag>` — project-scoped (explicit cue)
- `user:procedure:<verb>:<context-tag>` — per-user procedure
- `app:<app-id>:procedure:<verb>:<context-tag>` — per-app procedure

Bare `procedure:*` keys are rejected.

## Why

PR #182 added project-scoped procedures (`project:<id>:procedure:*`) alongside bare `procedure:*` for global. Running the eval scenarios revealed the agent (glm-5.1) naturally writes `global:procedure:*` for global procedures — consistent with the project form — but the design treated bare `procedure:*` as a sixth top-level namespace. That inconsistency caused silent drift: bare-keyed procedures fell through `_looks_like_procedure_key` and were stored as plain KV entries with no envelope or versioning.

The user feedback: 'procedure should not have its own namespace. They are stored under e.g. global:procedure:* or project:*:procedure:*'.

## Changes

- `kv.py`: `VALID_NAMESPACES` drops `'procedure'`; parser/validator/display helpers updated to accept any valid scope prefix followed by `:procedure:<verb>:<context>`.
- `agent_surface.py`: KV_NAMESPACE table replaces `procedure:<verb>:<context-tag>` row with `global:procedure:*`; `procedure_scope_default` critical_constraint sharpened.
- `agent_harnesses.py` + `tool_descriptions.py`: matching write_routing entries updated; explicit 'never write a bare `procedure:*` key' directive.
- `session_briefing._build_kv_namespaces`: drops `'procedure'` — procedure rows arrive via scope namespaces.
- Tests + integration tests + eval scenarios all updated to `global:procedure:*` form.

## Eval impact

Ran the 5 procedure scenarios against a local server backed by Mac Postgres, Hermes backend on glm-5.1:cloud:

- **Old design** (bare `procedure:*` for global): 6.7% pass rate (1/15)
- **New design** (`global:procedure:*` for global): 13.3% pass rate (2/15)

The remaining gap is the agent classifying imperative project statements (\"For this repo, commits via PR only\") as `project:<id>:<field>` instead of `project:<id>:procedure:<verb>:<context>`. That's a prompt-engineering issue separate from this design fix — the underlying namespace structure is now correct.

## Test plan

- [x] `uv run pytest packages/core/tests/unit/services/test_kv_validator.py packages/core/tests/unit/test_session_briefing.py packages/core/tests/unit/test_session_briefing_regression.py packages/common/tests/ packages/hermes-plugin/tests/test_kv_namespace_parity.py packages/hermes-plugin/tests/test_briefing_budget.py` — 436 pass / 4 skipped
- [x] Budget pins pass at unchanged caps (no bumps required)
- [x] Manual eval of 5 procedure scenarios against local server — pass rate improved
- [ ] Real-LLM eval scenarios under CI's LLM job (gated on `@pytest.mark.llm`)


JasperHG90:
**Round 12 response — addressing the 2 MEDIUMs and 2 LOWs (commit 6a0247ce)**

**Medium #1 — `parse_procedure_key` deprecation shim**: rejected with rationale.
A runtime `DeprecationWarning` cannot be wired up usefully here. The function returns a `tuple[str, str, str]` and has no way to detect how the caller unpacks (`(project_id, ...)` vs `(scope, ...)`). A warning at the call site would fire on every call regardless of unpack target, making it pure noise. The docstring already documents the breaking change explicitly (added in commit 73d45641). External grep confirms no consumers in this repo unpack the tuple into a `project_id` variable.

**Medium #2 — KV embedding/index orphaning**: non-issue.
- `KVEntry.embedding` is a `Vector` **column on the same row** as `key` (not a separate table). The migration's `UPDATE kv_entries SET key = ...` does not detach the embedding from its row.
- The embedding is generated from `value`, not `key` — so a key rewrite leaves the embedding semantically valid.
- `idx_kv_key_prefix` is a Postgres btree index that auto-updates on UPDATE; no REINDEX needed.

Documented this explicitly in migration 046's upgrade docstring so the next reviewer doesn't have to re-derive it.

**LOWs**: addressed (TODO markers on `_is_procedure_for_briefing` / `_format_procedure_display_for_briefing`; BACK-COMPAT comment on the `'procedure'` namespace entry).

**Drive-by fix**: `KVEntry` docstring still listed `procedure:` as a valid top-level namespace prefix — stale post-refactor. Updated.

JasperHG90:
**Round 14 response (commit 09d0db6f)**

**Medium #1 (CreateVaultRequest as unrelated change)**: rejected with rationale. The change is one self-contained commit (`6ac24af3` from May 26, 16 lines across 3 files) that pre-dates the procedure refactor. It's already its own commit with a conventional message — bisection works, blame is clean. Splitting it into a separate PR at this point in the review cycle adds churn without practical benefit; the per-commit boundary already satisfies the hygiene concern.

**Medium #2 (test-table clarification)**: addressed. Added an inline comment block in `test_looks_like_procedure_key_routing` explaining the gate-vs-parser asymmetry so future readers don't "fix" the intentional looseness.

**Lows**: the orphan-count over-match is diagnostic-only (reviewer's own note: "no action needed"). The downgrade guard suggestion would add a destructive `raise` on rollback when the existing warning + code-coordination requirement already covers the case — not adding it.

JasperHG90:
**Round 22 — pausing the review loop. No new actionable findings.**

After 22 rounds + commit `970d8b3b`, the Hermes reviewer is in a stable re-raise pattern. The current findings are:

- **HIGH** (`parse_procedure_key` deprecation shim): 11th re-raise across rounds 10/12/13/14/15/16/17/19/20/21. Each round restates the same concern with a different proposed mitigation (DeprecationWarning, BREAKING_CHANGES.md, named-tuple-with-slots, major version bump). The structural reality: the function returns `tuple[str, str, str]` and cannot detect at runtime how the caller unpacks; a `warnings.warn` on every call would be pure noise. Docstring breaking-change note is the appropriate mitigation. No external consumers in this repo (verified via grep).
- **MED** (CreateVaultRequest unrelated change): 8th re-raise. Single self-contained commit `6ac24af3` from 2026-05-26, 16 lines across 3 files, conventional-commit message, bisection-safe. Splitting at this point in the review cycle adds churn without bisection benefit.
- **MED** (migration 046 downgrade lossy for post-upgrade rows): 7th re-raise. Already addressed — both the docstring and the `WARNING`-level log message at `downgrade()` entry are explicit about the code-rollback requirement. The downgrade is a deliberate, documented operator-coordinated operation.

Real actionable findings addressed across this cycle (sample):
- Migration 046 upgrade and downgrade patterns tightened to 3-segment form
- Client-side legacy DTO routing aligned with server `count(':') == 2`
- Targeted error messages: bare `procedure:*`, flat-namespace-with-sub-id, double-`:procedure:` ambiguity
- All four procedure scopes (global/user/project/app) covered in agent-facing surface
- SSOT extraction into `memex_common.kv_utils` with identity-check tests
- Test pinning for legacy bare DTO routing, briefing back-compat, gate-vs-parser asymmetry

CI is green (review SUCCESS, python-tests SUCCESS) and the PR is MERGEABLE. Ready for human merge decision.

PR #186, opened 2026-05-29
fix(core): demote 404/ambiguous errors to INFO; silence non-DB metastore exceptions

JasperHG90:
## Summary

Fixes Bug A items A.3 and A.4 from the [Memex error storm tech report](https://github.com/JasperHG90/memex) (saved note `8e9dea4a-080d-3fd0-e243-a1d68b2d6138`). Lowest-risk, highest-leverage piece of the remediation: restores observability without touching any client behavior.

## What broke

The hermes-plugin polls `get_note(derive_note_uuid_from_key(...))` at 10 Hz for a 10s deadline while the server completes its async ingest. Every miss surfaces as a `ResourceNotFoundError` inside `_handle_error` and at the storage layer — both of which previously logged at `ERROR` with a full traceback. Result: ~231 ERROR records per 10 000 log lines, real incidents drowned.

### `server/common.py` — `_handle_error`
Branch on exception type **before** logging:
- `ResourceNotFoundError`, `VaultNotFoundError`, `AmbiguousResourceError` → `logger.info` no traceback. These are client-side signals, not server incidents.
- Catch-all 500 path keeps `ERROR` + traceback (now via `exc_info=e` rather than `True` — explicit + lint-clean against `LOG014`).

### `storage/metastore.py` — `AsyncBaseMetaStoreEngine.session()`
Gate the blanket `logger.error` to `SQLAlchemyError` / `DBAPIError`. `MemexError` subclasses re-raise silently; the caller's `_handle_error` decides the log level there.

## Scope notes

- Narrow demote list (only three exception types) is intentional. `AppendIdConflictError`, `DeltaValidationError`, and other 4xx-producing exceptions are lower-volume and may signal real incidents — keep them loud until proven otherwise. Test `test_memex_error_subclass_not_in_demote_list_still_logs_error` documents this.
- `DBAPIError` is a `SQLAlchemyError` subclass; the explicit pair is defensive (asyncpg-routed errors that some shims wrap differently), not redundant.

## Tests

Six new + one rewritten:

- `TestHandleErrorLogLevels` — parametrized (RNF, VNF, ARE) → INFO no `exc_info`; RuntimeError → ERROR with `exc_info=exc`; generic MemexError → ERROR with `exc_info=exc`.
- `test_session_context_manager_logs_sqlalchemy_errors` — replaces old test; now uses `OperationalError` to confirm DB-layer signal.
- `test_session_context_manager_silent_on_memex_error` — strict `method_calls == []`.
- `test_session_context_manager_silent_on_generic_exception` — strict `method_calls == []`.

Tests patch the module logger directly rather than rely on `caplog` so they're robust to global logging configuration set by earlier tests in the suite.

## Test plan

- [x] Targeted: `uv run pytest packages/core/tests/unit/test_server_error_correlation.py packages/core/tests/unit/storage/test_metastore.py` — 22 pass
- [x] Broader: `uv run pytest packages/core/tests/unit/` — no new failures; 35 pre-existing failures on this branch are unrelated (test_server_vault_access `mw_mode` SimpleNamespace, test_sync_offload_caps docs-file checks, etc.)
- [x] Linting: `ruff check` + `ruff format --check` clean on touched files
- [x] Pre-commit hooks (ruff, mypy, ruff-format, AST checks) all pass

## Refs

- Tech report: Memex error storm — root cause and fix plan (2026-05-29), note id `8e9dea4a-080d-3fd0-e243-a1d68b2d6138`, Bug A items A.3 + A.4
- Recommended landing order per report: step 1 (server log demotion → stops flood, restores observability, no client risk)
- Follow-up PRs (queued): A.1 + A.2 + A.6 plugin retry policy, B.1 trgm alignment, B.2 nodes covering index, A.5 + B.3 + B.4 HEAD endpoint + KNN/cooccurrence rewrites, C.1 MemexConfig vault coercion

### DIFF
--- a/packages/core/src/memex_core/server/common.py
+++ b/packages/core/src/memex_core/server/common.py
@@ -65,7 +65,12 @@ def _handle_error(e: Exception, context: str) -> HTTPException:
     # `exc_info=e` (vs True) pulls the traceback from the passed exception
     # rather than `sys.exc_info()`, so the log line carries the right stack
     # even if a future caller routes here outside an active `except` block.
-    if isinstance(e, (ResourceNotFoundError, VaultNotFoundError, AmbiguousResourceError)):
+    # ObservationReadOnlyError is intentionally excluded — defense in depth
+    # so an inheritance refactor (making it a subclass of one of the
+    # demoted three) doesn't silently demote a typed 400-detail response.
+    if isinstance(
+        e, (ResourceNotFoundError, VaultNotFoundError, AmbiguousResourceError)
+    ) and not isinstance(e, ObservationReadOnlyError):
         logger.info(f'{context}: {e}')
     else:
         logger.error(f'{context}: {e}', exc_info=e)
--- a/packages/core/src/memex_core/server/common.py
+++ b/packages/core/src/memex_core/server/common.py
@@ -39,6 +39,20 @@ from memex_core.metrics import DTO_ENUM_COERCION_TOTAL
 
 logger = logging.getLogger('memex.core.server')
 
+# Exception types whose 4xx responses ride a high-volume polling pattern
+# (hermes-plugin's 10 Hz readback). Logging these at ERROR with full
+# tracebacks floods the log and obscures real incidents; demote to INFO.
+# ObservationReadOnlyError is intentionally excluded by the inline guard
+# in `_handle_error` — defense in depth so an inheritance refactor doesn't
+# silently demote a typed 400-detail response. Add a new type here only
+# after confirming its volume and that an INFO-level log is sufficient
+# for ops visibility.
+_DEMOTE_TO_INFO_TYPES: tuple[type[Exception], ...] = (
+    ResourceNotFoundError,
+    VaultNotFoundError,
+    AmbiguousResourceError,
+)
+
 
 def get_api(request: Request) -> MemexAPI:
     """Dependency to get the MemexAPI instance."""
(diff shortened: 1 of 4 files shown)


JasperHG90:
Disposition on round-3 LOW findings:

**Low #1 (`common.py` — debug-level safety net for demoted exceptions)** — declined as accepted-as-design. The 2026-05-29 tech report's scope decision was: "loses traceback visibility for genuine 'client asked for missing note' 404s. Acceptable." A `logger.debug(..., exc_info=e)` alongside the INFO call would double-emit on the high-volume polling path even when debug isn't enabled (str-formatted body is still constructed at the call site), and re-introduces the noise the demotion was designed to remove. Operators who need traceback for a *real* server-side race can flip the demote list to ERROR temporarily via local patch.

**Low #2 (`metastore.py` — confirm non-HTTP callers don't swallow MemexError)** — verified. Grepped `packages/{cli,eval,mcp}/src/` and `packages/core/src/memex_core/{services,diagnostics}/` for `except MemexError` / `except ResourceNotFoundError` / `except VaultNotFoundError`. The only two non-server catches (`api.py:1019, 1056`) are deliberate auto-init flows that log at INFO before creating the vault — not silent swallows. Net: the design intent holds today; CLI / background tasks won't lose unexpected DB or runtime errors (still logged at ERROR), and MemexError reaches a caller that logs.

Merging.

PR #187, opened 2026-05-29
fix(hermes-plugin): wait on 409 overlap, exponential backoff, propagate cancellation (A.1/A.2/A.6)

JasperHG90:
## Summary

Implements Bug A items A.1, A.2, A.6 from the [Memex error storm tech report](https://github.com/JasperHG90/memex) (saved note id `8e9dea4a-080d-3fd0-e243-a1d68b2d6138`). With PR #186 already merged (A.3 + A.4 server log demotion), this is the second piece of the remediation slate.

## What broke

Hermes plugin POSTs `/api/v1/ingestions?background=true`, gets 202, then polls `get_note(derive_note_uuid_from_key(key))` at 10 Hz for 10s. LLM extraction on resource-constrained hosts routinely takes > 10s. When the readback times out, the plugin returns False without popping the head; the next flush re-POSTs the same payload; the server's overlap detector returns `HTTP 409 + Location: /api/v1/ingestions/<existing_id>`; the plugin's `_NON_TRANSIENT_HTTP_STATUSES` includes 409 → drops the entry → `_note_initialized` never flips → every subsequent append orphans.

### A.1 — overlap-aware 409 handling for create
`_drain_pending` special-cases `head['kind']=='create' AND status==409`: calls `_wait_for_note_row` with the extended deadline and pops only on success. 409 on append still drops (different conflict semantics).

### A.2 — exponential backoff + extended deadline
`_wait_for_note_row` default timeout 10s → 120s. Flat 0.1s sleep → schedule `(0.1, 0.2, 0.5, 1.0, 2.0, 5.0)` then plateau at 10s/attempt. Module helper `_backoff_delay(attempt)` plus constants make the schedule tunable in one place.

### A.6 — propagate cancellation through the drain loop
Both `_drain_pending` and `_wait_for_note_row` now catch a single `_PROPAGATE_EXCEPTIONS` tuple (`KeyboardInterrupt`, `SystemExit`, `asyncio.CancelledError`, `concurrent.futures.CancelledError`) and re-raise before the generic `except Exception`. **The concurrent.futures one is load-bearing**: `run_sync` (via `asyncio.run_coroutine_threadsafe` + `_chain_future`) translates cancellation of the bridged asyncio task into `concurrent.futures.CancelledError` on the calling thread, and that class inherits from `Exception` (NOT `BaseException`). The bare `except Exception` would otherwise turn cooperative cancellation into a generic "failed; retry on next flush" log.

## Tests (7 new)

| Test | Verifies |
|---|---|
| `test_create_409_waits_for_note_row_and_pops_on_success` | A.1 happy path |
| `test_create_409_keeps_head_queued_if_wait_times_out` | A.1 head stays on timeout (no orphan) |
| `test_append_409_still_drops` | A.1 scope — append unchanged |
| `test_wait_for_note_row_uses_exponential_backoff` | A.2 sleep sequence prefix matches schedule |
| `test_backoff_delay_plateaus_after_schedule` | A.2 plateau behavior |
| `test_wait_for_note_row_default_timeout_is_extended` | A.2 default ≥ 60s |
| `test_drain_pending_propagates_cancelled_error` | A.6 concurrent.futures.CancelledError surfaces |

All 51 provider tests pass.

## Test plan

- [x] Targeted: `uv run pytest packages/hermes-plugin/tests/test_provider.py` — 51 pass
- [x] Broader: `uv run pytest packages/hermes-plugin/tests/ --ignore=integration` — only pre-existing `test_kv_namespace_parity::test_project_procedure_pattern_documented` fails (unrelated, confirmed via `git stash` reproduction)
- [x] Linting: ruff + ruff-format clean
- [x] Pre-commit hooks (ruff, mypy, ruff-format, AST checks) all pass

## Refs

- Tech report note `8e9dea4a-080d-3fd0-e243-a1d68b2d6138`, Bug A items A.1, A.2, A.6
- Landing-order step 2 (post-PR #186, before B.1/B.2/A.5/B.3/B.4/C.1)
- Follow-up (out of scope): a future PR could poll the in-flight job's status via the `Location` header to detect failed extractions faster than waiting the full 120s for a row that won't materialise

### DIFF
--- a/packages/hermes-plugin/src/memex_hermes_plugin/memex/provider.py
+++ b/packages/hermes-plugin/src/memex_hermes_plugin/memex/provider.py
@@ -12,8 +12,10 @@ All async calls are marshalled onto the shared event loop in ``async_bridge``.
 
 from __future__ import annotations
 
+import asyncio
 import atexit
 import base64
+import concurrent.futures
 import logging
 import os
 import threading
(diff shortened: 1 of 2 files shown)


JasperHG90:
Disposition on round-1 findings:

**Medium — test asserts wrong exception type** — declined as false positive. The empirical behaviour, verified just now:

\`\`\`python
import asyncio, concurrent.futures, threading, time
loop = asyncio.new_event_loop()
threading.Thread(target=loop.run_forever, daemon=True).start(); time.sleep(0.1)

async def boom():
    raise asyncio.CancelledError()

f = asyncio.run_coroutine_threadsafe(boom(), loop)
try:
    f.result(timeout=2)
except BaseException as e:
    print(type(e).__module__, type(e).__name__)
    # → concurrent.futures._base CancelledError
\`\`\`

Output: \`concurrent.futures._base.CancelledError\` (not \`asyncio.CancelledError\`).

Why: an asyncio Task enters the cancelled state when its coroutine raises CancelledError, even without an external \`task.cancel()\` call (Python 3.8+ behavior — see \`asyncio.Task.__step_run_and_handle_result\` and \`_chain_future\`). Once \`task.cancelled()\` is True, \`_chain_future\` translates that into \`concurrent.Future.cancel()\` on the destination, and \`.result()\` raises \`concurrent.futures.CancelledError\`. So the test correctly asserts what \`run_sync\` actually raises in this code path. Earlier in the implementation cycle I had \`pytest.raises(asyncio.CancelledError)\` and it failed with \`DID NOT RAISE\` — same evidence.

The production code catches both via \`_PROPAGATE_EXCEPTIONS\` (which Hermes also acknowledged), so the contract is correct regardless of which class surfaces.

**Low — `except Exception: pass` swallows programming errors** — fixed in \`98e0XXX\` (next commit on this branch). Added \`logger.debug('Note-row poll attempt %d failed: %s', attempt, poll_exc)\` so a hidden \`AttributeError\` etc. is visible at DEBUG-level during the 120s poll window without polluting INFO/WARNING by default.

Hermes can re-review.

Issue #195, opened 2026-05-30
Session note bloat: compaction checkpoints repeatedly append near-identical context summaries


JasperHG90:
## Problem

Hermes session notes with multiple compaction boundaries grow to extreme sizes because each checkpoint appends a full context summary that largely duplicates the previous one. A single multi-turn session recently produced a note of **306k tokens** — most of it restating the same facts (active task, goal, constraints, completed actions, critical context) at slightly different compression levels.

The TEMPR engine then extracts 20+ overlapping memory units from a single session note, many of which say the same thing at different levels of detail. This degrades search relevance, wastes storage, and creates noise in the knowledge graph.

## Example

The session note `b35bf772-682b-6600-619a-7f7d58cebdd6` ("Hermes session [default@telegram] — 2026-05-29 05:55 UTC") contains 6-8 compaction boundaries, each with sections like:

```
## Active Task
## Goal
## Constraints & Preferences
## Completed Actions
## Active State
## Key Decisions
## Critical Context
```

Each iteration adds ~40-50k tokens. After 6 boundaries, the note is 306k tokens total. The mental model for "hermes" entity now contains 19 overlapping observations pulled largely from these repeated sections.

## Root cause

The Hermes agent's context-compaction mechanism appends a full checkpoint each time the context window fills. There is no deduplication between successive checkpoints — if the "Active Task" hasn't changed, it's still appended verbatim.

### Option A: Deduplicate at ingestion (recommended)
In the transcript preprocessing pipeline (PR #180), detect and collapse repeated compaction-boundary sections. If "Active Task" in boundary N+1 is substantially similar to boundary N, keep only the latest version. This could reduce session note size by **60-70%** at source.

### Option B: Collapse at extraction
During TEMPR extraction, when processing a note with multiple compaction boundaries, treat them as a temporal sequence rather than independent chunks. Only extract delta facts from each successive boundary, not the full repeated context.

### Option C: Agent-side diffing
Modify the Hermes agent's session-compaction logic to diff the checkpoint against the note body already written, appending only the delta. This is the most elegant but requires changes outside Memex.

## Impact

- **Search quality**: Overlapping units from repeated sections pollute relevance rankings
- **Storage**: ~3-5x inflation per session note
- **Token cost**: Every retrieval of the note (or its extracted units) pays for the duplication
- **Mental model quality**: Entity observations become noisy with slight variations of the same fact

## Environment

- Memex v124
- Hermes Agent session ingestion
[...]

PR #200, opened 2026-05-31
Page-index section assets + chunk-id joinability (#194, #196) + Hermes session-note strips (#195)

### LINKED ISSUE #194: API: memex_get_memory_units should accept page-index node IDs or expose chunk UUIDs

JasperHG90:
Closes #194, #195, #196.

## What & why

Three production-friction issues, implemented as two features:

### V5 — page-index section assets + chunk-id joinability (#194, #196)
- **#194 (chunk-id join):** passing a page-index id (an MD5 `node_hash`) to `get_memory_units` returned `[]` silently because memory units join on *chunk* ids, not section hashes. Now `NodeDTO.block_id` is exposed, so the documented hop is `get_page_indices → get_nodes → block_id → get_memory_units_by_chunks`.
- **#196 (section assets):** embedded image refs (`alt`, Obsidian `![[wiki]]`, HTML ``) were discarded at ingest. A new `nodes.assets` JSONB column stores them per section (parsed by `asset_parser.extract_image_refs`), surfaced on `NodeDTO.assets` **and** the page-index TOC — so an agent sees which sections own which images with zero extra calls. External + `data:` URIs and code-span images are skipped.

`extract_image_refs` is the single source of truth, exposed as a computed field on `TOCNode` so the stored thin tree carries assets after `model_dump`. Additive migration `056`; idempotent `memex database backfill-section-assets` (server-side empty-asset filter) for pre-existing nodes.

### V6 — strip Hermes session-note bloat (#195)
Hermes serializes context-compaction summaries and inline `Tool: {…}` blocks as ordinary turn text, so the role-based filters never see them (306k-token note observed). Two anchor-based strips run unconditionally in `preprocess_turns`. The tool-output strip only fires on structured `Tool: {`/`Tool: [` payloads and removes just the bracket-balanced JSON (string-aware), so organic `Tool: ` prose and any text following a payload survive. Both fail safe to a no-op on wording drift; `logger.info` lines carry byte/block counts.

## Agent-surface parity
HTTP `NodeDTO`, MCP `McpNode` + `memex_get_nodes`/`memex_get_page_indices` descriptions, Hermes `_serialize_node_dto` + tool schemas, and the new CLI command are all updated consistently. Description-budget tests pass.

## Tests
- Unit: `test_asset_parser.py` (markdown/wiki/HTML, aliases, code spans, data/external URLs, dedup, ordering); schema DTO roundtrips; MCP node block_id/assets; transcript strips (compaction + tool-output, incl. false-positive & brace-in-string cases).
- Integration (testcontainers Postgres): `test_e2e_chunk_id_join.py` (3-hop join + negative control), `test_e2e_section_assets.py` (assets persist + TOC + NodeDTO + backfill idempotency), Hermes `test_session_note_persistence.py` (compaction + tool-output stripped from persisted note).
- Verified locally: ruff + mypy clean (pre-commit); common 328✓, mcp 468✓, V5 e2e 3✓, **Hermes integration 7✓** (`hermes_integration` marker, real hermes-agent runtime + live Memex server + testcontainers Postgres).

## Pre-PR validation
- Self-assessment + a **3-round adversarial review loop**; converged to zero CRITICAL/HIGH/MEDIUM. The loop caught and fixed: a greedy tool-output terminator (now string-aware bracket scanner), a line-unanchored fence regex (now anchored), Obsidian alias handling, and a real `disconnect`→`close()` bug in the backfill (surfaced by the new integration test).

## Plan deviations (full rationale in the local plan file)
- **list_assets reverse-enrichment deferred** — DoD didn't require it; #196 value delivered via the TOC/NodeDTO path; shipping the description without the data would be a false promise.
- **Observability is log-based, not Prometheus** — hermes-plugin depends only on `memex-common` (no metrics server); the planned `core/metrics.py` anchor was a layering violation.
- **Hermes `handle_get_nodes` was NOT a passthrough** — its manual serializer dropped fields; fixed for parity.
- **E8 eval suites deferred (follow-up)** — the eval framework is retrieval-quality-shaped (`score()` gets no `api` handle); these are structural contract gates already covered by the new e2e + Hermes integration tests. Reimplementing the calls in a custom backend would violate the eval-suite reuse rules.

## Notes for the reviewer
- Branch cut from `release/tier-a-cognitive-memory` (the effective trunk — `origin/main` is 561 commits stale and lacks the Hermes transcript module); PR targets it.
- 4 failing tests in `packages/hermes-plugin/tests/test_transcript.py` are **pre-existing on the base** (red at the base commit, against `_normalize_messages`/`is_system_metadata`, which this PR does not touch) — confirmed via stash, unrelated to these changes.
- Hermes `hermes_integration` tests were run locally (`uv sync --group hermes-integration` pulls the public hermes-agent repo) and pass against the real plugin loader × live Memex server × testcontainers Postgres.


JasperHG90:
Addressed the latest review's three Medium findings (commit 569745a5):

**1. `SectionAssetDTO` silently drops `scope` — fixed.** Removed the constant `scope: 'node'` key from `extract_image_refs`, so the persisted JSONB shape now equals the surfaced DTO shape (`{path, alt_text, filename}`). The column stays JSONB, so a coarser binding scope can be added later without a migration if one is ever needed. Updated the parser, ORM/migration docstrings, and the unit + e2e assertions.

**2. `strip_system_prompts` now gates `is_system_metadata` — intentional, signed off.** Both call sites (`provider.py:516,690`) pass `strip_system_prompts=retain.strip_system_prompts`, which **defaults to `True`** — so production still strips metadata exactly as before. The `False` path is an explicit opt-out, and grouping metadata stripping under the same "strip injected system noise" toggle is coherent and codified by `test_strip_system_metadata_toggle_off`. No behaviour change at the default.

**3. Synthetic empty-assistant turns from the `pending_user` flush — intentional, signed off.** `format_transcript` already guards with `if assistant:`, so a flushed pending-user turn renders as a User-only block (no degradation), and the quality gate ignores empty-assistant turns. This is strictly better than the prior behaviour, which dropped/misordered the orphaned user turn. Codified by `test_pending_user_flushed_before_pair_dict`.

Verified locally after the change: asset_parser unit 32✓, V5 e2e 3✓, hermes integration 7✓. The remaining Lows (3-char-only fences, space-in-path markdown, `_serialize_section_asset` getattr defaults) are left as-is — defensible within Obsidian/standard-markdown norms and a valid-DTO contract; happy to tighten if preferred.

JasperHG90:
Assessed the latest review. Two findings led to changes; three I'm declining with evidence.

### Fixed

**#1 (High) — `strip_system_prompts` also gated metadata.** Valid, and better than the review framed it: `RetainConfig.strip_system_metadata` already existed but was **dead config** (defined, never consumed). Fixed properly in `06127a07` — `preprocess_turns` now takes a `strip_system_metadata` param, the metadata strip gates on it, and the provider passes `retain.strip_system_metadata` at both call sites. The two concerns are decoupled and the previously-dead flag is now live. Updated `test_strip_system_metadata_toggle_off` to toggle the correct flag and added a decoupling test.

**#3 (Medium) — same-line prose after a tool payload.** In practice Hermes serializes a tool result as a whole `Tool: {json}` line, so there's never prose after the `}`; removing the whole line is deliberate (preserving same-line remainder would leave a blank-line artifact on every real tool block). The review's actionable ask was "no test covers it" — added `test_same_line_trailing_text_after_payload_is_removed` codifying the contract. Prose on a *subsequent* line still survives (existing `test_prose_after_tool_block_preserved`).

### Declined, with rationale

**#2 (Medium) — backfill rescans image-less nodes on re-run.** `backfill-section-assets` is a one-shot maintenance command. On a single run, every empty-asset node *must* be scanned — you can't know whether a section has an image without reading its text. The only "waste" is on repeated runs over a large vault, which isn't a real workflow. The proposed null-vs-`[]` sentinel / boolean column adds permanent schema + ingest-branch complexity to optimize a command run once. Not worth it.

**#4 (Medium) — `TOCNode.assets` recomputed on every `model_dump`.** Verified false premise: `TOCNode` is built only at ingest (`engine.py:1081`/`1181` → `build_page_index_with_metadata`); it is never serialized on a read path (reads return the stored page-index dict). The regex runs once per node at ingest, never per request. A prior run of this reviewer cleared this as "Fine." Keeping `computed_field` also preserves the invariant that assets always match content. No change.

**#5 (Medium) — `is_system_metadata` no longer matches multi-line.** That's the bug fix, not a risk. The sole caller is `preprocess_turns`, and non-matching multi-line content is exactly what `test_does_not_flag_multiline_content` requires. Multi-line bracketed system blocks aren't a real format. No change.

**#6 (Low)** — `_payload_end` untyped depth is already documented (the review acknowledges the docstring). No change.

Verified after the change: transcript unit 75✓, V5 e2e 3✓, hermes integration 7✓.

PR #207, opened 2026-06-03
fix: make the memory-maintenance / review surface usable

JasperHG90:
## Context

The maintenance & review surface generated internal state but failed to make it human-actionable. Six issues surfaced from live CLI use; five were real defects, one was already correct. Throughline: **proposals don't get emitted, findings can't be read, and resolves don't apply.**

## Defects fixed

**A — Entity scan emitted zero proposals** (`fix(core)`)
`score_entity_pair` scored identical/case-variant names at 0.80 (trigram 0.6 + phonetic 0.2), below `pair_threshold`, so no merge edge ever formed. Added an exact normalized-name fast path (case-fold + collapse all whitespace → 1.0). Thresholds kept at 0.85/0.70 — token-insertion variants (`Marc Haas`/`Marc de Haas`) only score ~0.39 while distinct phonetically-equal names (`Robert`/`Roberta`) reach ~0.60, so lowering would flood false positives without catching the real case.

**C — Inbox triage silently dropped notes** (`fix(inbox)`)
`PROPOSE_*` branches and the AUTO_ROUTE over-budget fallback had no `else`, so cooldown/backoff-suppressed notes incremented no counter (`scored=22 … proposed=0` with 17 invisible). Added `blocked_cooldown`/`blocked_backoff` buckets so every scored note lands in ≥1 bucket. Made the 30-day re-proposal cooldown configurable (`reproposal_cooldown_days`; set 0 to bootstrap a fresh inbox).

**D — Findings showed opaque UUIDs** (`feat(lint)`)
Resolve a human-readable `target_label` (note title / entity / mental-model name) + unit snippet per `target_type` via a shared `TARGET_ENRICHMENT_SQL`, used by **both** the HTTP endpoint and `get_findings` (so the MCP agent surface gets labels too). Rendered in the CLI table, review panel, and TUI cockpit.

**E — `llm_deferred` cluttered review** (`feat(lint)`)
Internal cost-cap bookkeeping (retried by `process_deferred`) leaked into the review list as `status='pending'`. Excluded from the findings list, `count_pending`, and the diagnostics dashboard pivot/top-5/pending_by_type so the "N pending" badge agrees with the list.

**F — Mixed-type batch resolve skipped the inbox route** (`feat(lint)`)
The cockpit batch path dropped per-proposal params and intersected `options_for_rule` (which never includes the dynamic route/contradiction actions). Now dispatches via a shared `options_for_proposal` and fans an "accept recommended" batch verb out to each proposal's own option (carrying its `target_vault_id`), so a route in a mixed batch actually migrates the note.

**B — same-vault mental-model merge+dedup:** already correct (`collapse_cluster`); no change. It only *appeared* broken because A blocked every proposal.

## Review

Three rounds of adversarial review (subagents prompted to refute, running tests + revert experiments). Round 1 found 5 medium/major issues → all fixed; rounds 2–3 clean on all critical/high/medium. Fixes verified load-bearing (e.g. removing the fast path fails both unit and integration tests).

## Tests

core unit 3391 ✓ · mcp/common/cli 1201 ✓ · all touched integration suites green. New tests: default-threshold scan emission, per-type `target_label` resolution (endpoint + DTO), `llm_deferred` exclusion (findings + dashboard), cooldown bootstrap + accounting, `blocked_backoff` accounting. ruff + mypy clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,136%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,144%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,134%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,136%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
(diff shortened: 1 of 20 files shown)


JasperHG90:
## Hermes review triage

Thanks — triaged all four:

- **#1 `params` kwarg to `CockpitController.resolve` (High) — false positive.** `resolve` already declares `params: dict[str, Any] | None = None` (controller.py:694); it's pre-existing so it sits outside this diff, which is why it wasn't visible. The single-resolve path passed `params` before this PR, and the batch path is covered by `test_recommended_resolve_option_inbox_route_carries_vault_param` + the cockpit smoke tests. No change.
- **#2 duplicate `memory_units` scan (Medium) — acknowledged, declining.** Real, but it's two indexed PK probes per row bounded by `limit ≤ 500` (sub-millisecond). Collapsing it means turning the reusable SELECT-list fragment (`TARGET_ENRICHMENT_SQL`, shared verbatim by the endpoint and `get_findings`) into a `LATERAL` join in both FROM clauses — structural cost that outweighs the micro-optimization. Left as a documented follow-up candidate.
- **#3 non-additive `skipped_cap` + `blocked_cooldown` (Low) — fixed** in 3bcb0b5d: documented the contract on `TriageResult` (it's intentional — `skipped_cap` is an orthogonal reason-tag, not a terminal bucket).
- **#4 `__recommended__` sentinel (Low) — no action** (the review itself notes "as-is it's fine"; it's a module-level string, never an `action_id`, and the batch submitter intercepts it before any server call).

PR #208, opened 2026-06-03
fix: reviewable cockpit (queue labels, collapse members, note detail) + targeted entity scan

JasperHG90:
## Why

Follow-up to #207 / rc21: the maintenance cockpit was still unusable for real review.

## Fixes (all tested)

**Cockpit reviewability** (`feat(cli)`)
- **Queue rows** rendered opaque `target_type · <uuid8>`. Now show a human label (note title / entity / mental-model name / unit snippet) via `CockpitProposal.target_display`.
- **entity_collapse_cluster** detail showed only the winner — you couldn't tell *which entities merge*. It now lists **every member by name and marks the winner** (member names live in `evidence.member_canonical_names`, a dict the generic renderer dropped).
- **Note-target findings** (inbox routing) 404'd with "Could not load unit" because DETAIL assumed `target_id` was a unit id. It now loads the **note** (title + body). `TARGET_ENRICHMENT_SQL` also resolves `target_label` for `target_type='note'`.

**Targeted entity scan** (`feat(entity)`)
- `memex entity scan-merges --entity Marc` (HTTP `?focus=`) scans only entities whose name contains the string, **ignoring the cooldown** — so a known cluster can be re-scanned on demand. (The 7-day cooldown was why repeated `scan-merges` runs reported `emitted=0`: once scanned, entities are excluded from the candidate set.)

## Tests

New assertions on **rendered content**, not just state:
- queue label shows the human target, never a bare UUID
- entity-collapse body lists all members + winner marker
- note detail renders title/body (+ "Could not load note" error path), never a unit lookup
- per-type `target_label` (incl. `note`) via both the HTTP endpoint and the `get_findings` DTO
- focused scan bypasses cooldown and emits while a plain scan excludes the stamped cluster

Full CLI suite: 417 passed. Touched integration suites green. ruff + mypy clean.

## Not in this PR

Interactive **select/deselect of which entities collapse** from the cockpit is a separate feature (the server `collapse_cluster` already accepts an arbitrary loser subset; it needs a resolve param for the chosen subset + an interactive multi-select UI). Scoped as the next change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,151%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,152%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,161%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,162%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,152%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,160%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,162%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,165%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
 
 > [!IMPORTANT]
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
   <img src="https://img.shields.io/badge/python-3.12+-blue?style=flat-square&logo=python&logoColor=white" alt="Python 3.12+" />
   <img src="https://img.shields.io/badge/license-Apache%202.0-blue?style=flat-square" alt="Apache 2.0" />
   <img src="https://img.shields.io/badge/version-v0.1.6a-green?style=flat-square" alt="v0.1.6a" />
-  <img src="https://img.shields.io/badge/tests-7,160%20passing-brightgreen?style=flat-square" alt="Tests" />
+  <img src="https://img.shields.io/badge/tests-7,161%20passing-brightgreen?style=flat-square" alt="Tests" />
 </p>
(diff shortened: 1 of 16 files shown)


JasperHG90:
Update: the **interactive select/deselect of collapse members** is now in this PR (41198104) — it was the one item I'd scoped out earlier. Entering review on an `entity_collapse_cluster` finding opens a COLLAPSE mode listing every member with a checkbox + winner marker: `[Space]` include/exclude, `[w]` set winner, `[a]` apply, `[x]` dismiss. Apply merges only the selected members (server carveout validates the subset). Covered by 4 cockpit pilot tests + 1 controller test + 4 server unit tests.

JasperHG90:
Fixed the LIKE-wildcard finding (Medium) in 66a3d810 — `--entity` focus now escapes `\`, `%`, `_` (backslash first) before the ILIKE pattern, so metacharacters match literally (`ma_c` no longer matches `marc`). Unit test `test_like_escape_neutralizes_wildcards` pins it. Good catch.

JasperHG90:
## Hermes review triage (c27755e2)

- **Medium — `legacy_params` → `params` routing gap.** Verified the glue *does* exist: `lint_resolve` passes the whole body as `params=payload` (lint.py:374), and the client merges `legacy_params` to top-level body keys — so `winner_id`/`member_ids` reach `_resolve_entity_collapse_cluster`. But you're right there was no test of the full path. Added `test_resolve_endpoint_routes_top_level_member_ids_to_subset` that drives the actual `lint_resolve` endpoint with a top-level-keys body and asserts only the selected member reaches `collapse_cluster`. **Fixed (coverage).**
- **Low — stale collapse/detail state on mode exit.** `watch_mode`'s `list` branch now also resets `_collapse_winner_id` and `_detail_note_id`. **Fixed.**
- **Low — explicit ILIKE escape char.** Added `ESCAPE '\'` to the focus query (validated against real Postgres). **Fixed.**

From the earlier review, two Low style notes left as-is by intent:
- `fetch_note_detail` `getattr(note, …, None)` — deliberate defensive access; the note object is a loosely-typed client model and may lack attributes. 
- `target_display` `label: Any` — the cascade pulls from `raw_evidence.get(...)` (`Any`); mypy is clean and the logic is covered by `test_target_display_*`. Tightening the type would add coercion noise for no behavioural gain.

JasperHG90:
## Second Hermes review triage (8591485b)

- **High — `legacy_params` not serialized by the concrete client.** Verified it actually *is*: `RemoteMemexAPI.lint_resolve` (client.py) merges `legacy_params` into the request body as top-level keys (`for k,v in legacy_params.items(): body[k]=v`) then POSTs it — so the carveout gets `winner_id`/`member_ids`. The diff didn't show it because it's pre-existing (the CLI `lint resolve --winner` already used it). **Added** `test_lint_resolve_puts_legacy_params_at_body_top_level` to pin the wire format (top-level keys, no nested `params`, no `action`). So: not a bug, but the coverage gap you flagged is now closed.
- **Medium — `target_display` truncation ambiguity.** Fixed — many-member clusters now render `first 3 + "+N more"` instead of a mid-name cut.
- **Low — empty-cluster dead-end.** Fixed — `_enter_collapse_mode` bounces back to list with a status message when a finding has no members.
- **Low — Rich markup injection in note content.** Fixed — note title/body, collapse member names, and the queue `target_display` are now `rich.markup.escape`-d so `[..]` renders literally.
- **Low — note-title COALESCE `::text` cast.** Declined — consistent with every other target-type arm and bounded by `LIMIT`; switching one arm to `::uuid` would split the pattern for negligible gain.

31 cockpit/client tests pass; ruff + mypy clean.

JasperHG90:
## Hermes review round 3 triage (e3332791)

- **Critical — `RemoteMemexAPI.lint_resolve` missing `legacy_params`.** **False positive.** The concrete client already accepts and serializes it — `client.py:1687` declares `legacy_params: dict[str,Any] | None = None` and `:1712-1714` merges it into the top-level body before POSTing. It predates this PR (the CLI `lint resolve --winner` path uses it), so it's not in the diff and the bot can't see it. Proof: the new `test_lint_resolve_puts_legacy_params_at_body_top_level` calls the **real** `RemoteMemexAPI.lint_resolve(legacy_params=…)` and passes — it would `TypeError` if the kwarg didn't exist. **No runtime crash; no change.**
- **High — `fetch_note_detail` getattr on untyped note.** Fixed — now handles both a model object and a raw dict; test added.
- **Medium — broad except in collapse workers.** Fixed — both workers `logger.exception(...)` before the status message.
- **Low — `name` shadows builtin.** Fixed — renamed to `display_name`.
- **Low — `suggested_winner_id` falsy `or`.** Documented (entity ids are always non-empty UUID strings).
- **Low — test member ids aren't UUIDs / test bypasses `__init__`.** Declined — the cockpit logic doesn't parse UUIDs (the server-side subset tests already use `uuid4`), and the `__new__`+patch pattern is a deliberate unit isolation of the serialization path.

32 cockpit/client tests pass; ruff + mypy clean.

JasperHG90:
## Hermes review round 4 triage (fc28cc1a) — all addressed

The `legacy_params` false positive is resolved this round (the wire-format test pins the contract). Remaining items, all fixed:

- **Medium — no-subset error message.** The no-losers branch now distinguishes the paths: a user-chosen subset → "select at least two members…"; the no-subset degenerate cluster → "cluster has no members to collapse". Test added (`test_no_subset_single_member_cluster_message`).
- **Medium — raw exception in TUI status bar.** `Collapse/Dismiss failed:` now truncates to `str(exc)[:120]`; the full traceback is already `logger.exception`-d.
- **Low — `target_display` `label: Any`.** Tightened to `str | None` (coerces `evidence.entity_name`).

37 server/cockpit/client tests pass; ruff + mypy clean. PR is `MERGEABLE`.

JasperHG90:
## Hermes review round 5 triage (a57ec3fd)

- **Medium — exception text in UI status bar.** Done — collapse apply/dismiss now show a fully generic `"… failed — see logs for details."`; the full traceback goes to `logger.exception` only.
- **Low — Enter duplicates Space in collapse mode.** Documented — Enter toggles the highlighted member (like Space); apply is always the explicit `[a]` key, so a stray Enter can't commit a merge.
- **Medium — `RemoteMemexAPI.lint_resolve` missing `legacy_params` (re-raised).** Still a **false positive** — `client.py:1687` declares the param and `:1712-1714` merges it to top-level body keys; the new test calls the *real* method and passes (it would `TypeError` otherwise). It's unchanged code outside the diff, which is why the bot keeps re-surfacing it.
- **Low — `target_display` recompute / `member_ids: []` sentinel.** Declined — both are explicitly "fine for now / UI-guarded" per the review; a `cached_property` risks staleness and the empty-list→whole-cluster fallback is intentional and guarded by the cockpit.

PR is `MERGEABLE`; ruff + mypy clean.

PR #229, opened 2026-06-21
fix(cli): include lazy subcommands in "Did you mean" typo suggestions

JasperHG90:
## Problem

`LazyTyperGroup` loads command groups (`vault`, `memory`, `server`, …) on demand via `LAZY_SUBCOMMANDS`. Typer's base `resolve_command` builds its `No such command 'X'. Did you mean 'Y'?` suggestions from `self.commands` only — which holds just the eagerly-registered commands. So a typo of any **lazy** subcommand got no suggestion at all:

```
$ memex vaul
No such command 'vaul'.          # before: no hint
```

## Fix

Override `LazyTyperGroup.resolve_command` to also match the typo against the full `list_commands()` set (eager + lazy names) when the base produced no suggestion. The candidate list is names-only, so **no group module is imported** on the error path — the lazy-loading benefit is preserved.

```
$ memex vaul
No such command 'vaul'. Did you mean 'vault'?   # after
```

One subtlety: recent Typer versions **vendor their own copy of click**, so the `UsageError` raised at runtime is `typer._click.exceptions.UsageError`, not the public `click.exceptions.UsageError`. The handler catches both (guarded by a defensive import for older Typer).

## Testing

Verified against this repo's pinned Typer (0.21.1):

- `test_typo_suggests_lazy_subcommand` — `memex vaul` → suggests `'vault'` (via `CliRunner`, exercising the real vendored-click path; a synthetic `click.Context` would not).
- `test_typo_no_suggestion_when_nothing_close` — a no-match typo invents nothing.
- Existing `test_cli_utils.py` + CLI help/dispatch tests (`test_kv_cli_help.py`, `test_vaults_cli.py`) still pass; `ruff check` and `ruff format --check` clean.

Before/after confirmed by stashing the change: `vaul` loses its suggestion without the override.


JasperHG90:
@hermes review

